PR #12, opened 2017-07-19
Add Count, Image fields to Spec

### LINKED ISSUE #9: Create habitat service CRD spec
This issue is about creating a CRD spec for a Habitat service.

Fields:

- Count
- Docker image

asymmetric:
This PR adds the `Count` and `Image` fields to the spec, representing the amount of instances, and the docker image to run.

Additionally, it removes a hyphen from the pluralization of the ServiceGroup CR, used as part of the k8s group identifier. That's because official k8s groups don't have hyphens between words (e.g. `apiextensions`).

And finally, it changes the group name for the CR, to resemble the `k8s.io` format used by the official components.

Closes #9.

### DIFF
--- a/pkg/habitat/apis/cr/v1/register.go
+++ b/pkg/habitat/apis/cr/v1/register.go
@@ -26,7 +26,7 @@ var (
 )
 
 // GroupName is the group name used in this package.
-const GroupName = "cr.habitat-operator.github.com"
+const GroupName = "habitat.sh"
 
 // SchemeGroupVersion is the group version used to register these objects.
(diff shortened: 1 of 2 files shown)


pkg/habitat/apis/cr/v1/register.go
```diff
@@ -26,7 +26,7 @@ var (
 )
 
 // GroupName is the group name used in this package.
-const GroupName = "cr.habitat-operator.github.com"
+const GroupName = "habitat-operator.kinvolk.io"
```
LiliC: I would say we go with `habitat.kinvolk.io`, or even `habitat.chef.io`, it makes more sense imho.

asymmetric:
PTALA.

PR #13, opened 2017-07-19
Turn the operator into a long-running process

asymmetric:
After starting the goroutines that listen for events, the operator's main thread simply waits for cancellation in the form of `SIGTERM` or of a message on the go [`Context`](https://golang.org/pkg/context/)'s channel.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -55,13 +55,17 @@ func run() int {
 	}
 
 	// Create ServiceGroup CRD.
-	crd, err := habitatclient.CreateCRD(apiextensionsclientset)
-	if err != nil && !apierrors.IsAlreadyExists(err) {
-		logger.Log("error", err)
-		return 1
-	}
+	_, crdErr := habitatclient.CreateCRD(apiextensionsclientset)
+	if crdErr != nil {
+		if !apierrors.IsAlreadyExists(crdErr) {
+			logger.Log("error", crdErr)
+			return 1
+		}
 
-	logger.Log("info", "created ServiceGroup CRD")
+		logger.Log("info", "ServiceGroup CRD already exists, continuing")
+	} else {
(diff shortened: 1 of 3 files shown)


pkg/habitat/client/cr.go
```diff
@@ -29,7 +29,11 @@ import (
 	"k8s.io/client-go/rest"
 )
 
-const serviceGroupCRDName = crv1.ServiceGroupResourcePlural + "." + crv1.GroupName
+const (
+	serviceGroupCRDName = crv1.ServiceGroupResourcePlural + "." + crv1.GroupName
+	pollInterval        = 500 * time.Millisecond
```
LiliC: Any reason for the time length you chose here? Just curious.

cmd/operator/main.go
```diff
@@ -55,41 +55,44 @@ func run() int {
...
-	}()
-
 	client, scheme, err := habitatclient.NewClient(config)
 	if err != nil {
 		logger.Log("error", err)
 		return 1
 	}
 
-	controller := habitatcontroller.HabitatController{
+	hc := habitatcontroller.HabitatController{
```
LiliC: Not a huge fan of the namings, since this is a habitat operator, naming things habitat**** is obsolete. wdyt?
But tbh can't come up with anything that makes sense right now so I guess we can leave it.

(I understand you want to differentiate from k8s controller for example, just does not look right if you know what I mean. :) )

pkg/habitat/client/cr.go
asymmetric: Completely arbitary but also coming from the example implementation.

cmd/operator/main.go
asymmetric: True, but OTOH I was following the [official example](https://github.com/kubernetes/apiextensions-apiserver/blob/master/examples/client-go/main.go#L69).

PR #17, opened 2017-07-20
Create Deployment in the onAdd handler

### LINKED ISSUE #10: Create deployment `onAdd`
Once the operator has received a new CR, it needs to create a Deployment using the CR's parameters.

asymmetric:
We create a Deployment with label `habitat=true` in the default namespace. The image and count are taken from the CO.

We introduce the `Topology` field to the spec, matching those supported in habitat.

Basic validation is performed on the CO, to ensure that a leader-follower Service Group has at least 3 services.

Closes #10.

### DIFF
--- a/pkg/habitat/apis/cr/v1/types.go
+++ b/pkg/habitat/apis/cr/v1/types.go
@@ -31,7 +31,8 @@ type ServiceGroupSpec struct {
 	// Count is the amount of Services to start in this Service Group.
 	Count int `json:"count"`
 	// Image is the Docker image of the Habitat Service.
-	Image string `json:"image"`
+	Image    string `json:"image"`
+	Topology `json:"topology"`
 }
 
 type ServiceGroupStatus struct {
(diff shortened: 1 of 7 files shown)


pkg/habitat/apis/cr/v1/types.go
```diff
@@ -31,7 +31,8 @@ type ServiceGroupSpec struct {
 	// Count is the amount of Services to start in this Service Group.
 	Count int `json:"count"`
 	// Image is the Docker image of the Habitat Service.
-	Image string `json:"image"`
+	Image    string `json:"image"`
+	Topology `json:"topology"`
```
robertgzr: Did you forget the type here?

asymmetric: Embedding FTW: `If we need to refer to an embedded field directly, the type name of the field, ignoring the package qualifier, serves as a field name`.

PR #19, opened 2017-07-21
Implement onDelete handler

asymmetric:
It deletes the deployment created in the `onAdd`.

Also, we add checks around the type conversions of the events we receive from the `Informer`.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -177,7 +177,7 @@ func (hc *HabitatController) onDelete(obj interface{}) {
 	level.Debug(hc.logger).Log("function", "onDelete", "msg", sg.ObjectMeta.SelfLink)
 
 	deploymentsClient := hc.config.KubernetesClient.Deployments(sg.ObjectMeta.Namespace)
-	deploymentName := fmt.Sprintf("%s-deployment", sg.Name)
+	deploymentName := sg.Name
 	deletePolicy := metav1.DeletePropagationForeground
 	deleteOptions := &metav1.DeleteOptions{
 		PropagationPolicy: &deletePolicy,
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -102,7 +102,12 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) (cache.Co
 }
 
 func (hc *HabitatController) onAdd(obj interface{}) {
-	sg := obj.(*crv1.ServiceGroup)
+	sg, ok := obj.(*crv1.ServiceGroup)
+	if !ok {
+		level.Error(hc.logger).Log("msg", "unknown event type")
+		return
+	}
+
 	level.Debug(hc.logger).Log("function", "onAdd", "msg", sg.ObjectMeta.SelfLink)
 
 	// Validate object.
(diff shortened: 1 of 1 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -163,6 +168,26 @@ func (hc *HabitatController) onUpdate(oldObj, newObj interface{}) {
...
-	level.Info(hc.logger).Log("function", "onDelete", "msg", sg.ObjectMeta.SelfLink)
+	sg, ok := obj.(*crv1.ServiceGroup)
+	if !ok {
+		level.Error(hc.logger).Log("msg", "unknown event type")
+		return
+	}
+
+	level.Debug(hc.logger).Log("function", "onDelete", "msg", sg.ObjectMeta.SelfLink)
+
+	deploymentsClient := hc.config.KubernetesClient.Deployments(apiv1.NamespaceDefault)
```
LiliC: Maybe we should follow the golang naming conventions and shorten these to `deploymentsClient ` -> `dc`, and for all the below vars. Like you did with `ServiceGroup`. Wdyt?

LiliC: We should also get the namespace from the `sg.ObjectMeta.Namespace` instead of hardcoding it to `default`.

pkg/habitat/controller/controller.go
```diff
@@ -163,6 +168,26 @@ func (hc *HabitatController) onUpdate(oldObj, newObj interface{}) {
...
+	deploymentName := fmt.Sprintf("%s-deployment", sg.Name)
```
LiliC: Do we need suffix this as deployment, we already know in k8s world that this is a deployment.

pkg/habitat/controller/controller.go
```diff
@@ -163,6 +168,26 @@ func (hc *HabitatController) onUpdate(oldObj, newObj interface{}) {
...
+	deletePolicy := metav1.DeletePropagationForeground
+	deleteOptions := &metav1.DeleteOptions{
+		PropagationPolicy: &deletePolicy,
```
LiliC: Why redeclare `DeletePropagationForeground`, we can just use that here.

pkg/habitat/controller/controller.go
asymmetric: Regarding the first comment, I agree that it's normally best practice, but in this case I don't know if it's clear enough that you're getting a client for deployments. @krnowak?

pkg/habitat/controller/controller.go
asymmetric: Because `PropagationPolicy` is a `*DeletionPropagation`, and `&metav1.DeleteOptions` doesn't compile.

LiliC: You are right, just noticed its a constant. 👍

pkg/habitat/controller/controller.go
asymmetric: Also I'm following the style in [the example](https://github.com/kubernetes/client-go/blob/master/examples/create-update-delete-deployment/main.go#L51).

krnowak: I don't know where these naming conventions for variables are (just skimmed the effective go document, but there's nothing about it I think), but I prefer to have them legible and unambiguous. Because `dc` is "drawing context" or "device context" for me. ;)

On the other hand, note that you also use `sq` for service group variable, so there's not much consistency here.

asymmetric: I consider long variable names useful to disambiguate.

In the case of `sg, ok := obj.(*crv1.ServiceGroup)`, there is no ambiguity as to what you're getting.

So for me, consistency is important, but clarity is more important.

@LiliC wdyt, can we merge this?

Issue #11, opened 2017-07-18
Write pod IP in the already created ConfigMap


LiliC:
Once the deployment is created we want to write the IP of the first running pod into a ConfigMap.

Relates to https://github.com/habitat-sh/habitat/issues/2735 `--peer-watch-file` flag that has to be implemented on the `hab` client.

PR #28, opened 2017-07-26
Create ConfigMap with peer IP file

### LINKED ISSUE #21: Create ConfigMap and mount it in the pod
First step to getting the full functionality of the habitat leader-follower topology is to create the ConfigMap, which we will later update with the pod IP. This will be taken by habitat when using the `--peer-watch-file` flag. See https://github.com/habitat-sh/habitat/issues/2735.

asymmetric:
This creates a ConfigMap, links it to the Deployment (so that deleting the Deployment deletes the CM), and mounts a file in the containers under `/kuberntes-configmap/peer-watch-file`.

The weird path is because the mount point must be reasonably unique, otherwise it will hide resources already present in the container's filesystem.

For now, the file is empty. It will be populated as part of #11.

The file(system) is mounted as readonly. In case we decide to allow supervisors to change it due to gossip events, we can mark it as read-write later on.

Closes #21.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -86,11 +86,16 @@ func run() int {
 	}
 
 	controllerConfig := habitatcontroller.Config{
-		HabitatClient:    habitatClient,
-		KubernetesClient: clientset.AppsV1beta1Client,
-		Scheme:           scheme,
+		AppsV1beta1Client: clientset.AppsV1beta1Client,
+		CoreV1Client:      clientset.CoreV1Client,
+		HabitatClient:     habitatClient,
+		Scheme:            scheme,
 	}
-	hc := habitatcontroller.New(controllerConfig, log.With(logger, "component", "controller"))
+	hc, err := habitatcontroller.New(controllerConfig, log.With(logger, "component", "controller"))
+	if err != nil {
+		level.Error(logger).Log("msg", err)
+		return 1
+	}
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
 	defer cancelFunc()
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -86,10 +86,9 @@ func run() int {
 	}
 
 	controllerConfig := habitatcontroller.Config{
-		AppsV1beta1Client: clientset.AppsV1beta1Client,
-		CoreV1Client:      clientset.CoreV1Client,
-		HabitatClient:     habitatClient,
-		Scheme:            scheme,
+		HabitatClient:       habitatClient,
+		KubernetesClientset: clientset,
+		Scheme:              scheme,
 	}
 	hc, err := habitatcontroller.New(controllerConfig, log.With(logger, "component", "controller"))
 	if err != nil {
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -27,8 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/kubernetes/typed/apps/v1beta1"
-	"k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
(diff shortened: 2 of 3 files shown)


pkg/habitat/controller/utils.go
```diff
@@ -40,6 +40,7 @@ func validateCustomObject(sg crv1.ServiceGroup) error {
 			return validationError{msg: "too few instances", Key: "count"}
 		}
 	default:
+		return validationError{msg: "unknown topology", Key: "topology"}
```
LiliC: Not sure you meant this to go in this PR?

cmd/operator/main.go
```diff
@@ -86,11 +86,16 @@ func run() int {
 	}
 
 	controllerConfig := habitatcontroller.Config{
-		HabitatClient:    habitatClient,
-		KubernetesClient: clientset.AppsV1beta1Client,
-		Scheme:           scheme,
+		AppsV1beta1Client: clientset.AppsV1beta1Client,
```
LiliC: Why not just pass in clientset? Or else we will keep adding new parameters for every client that we need.

pkg/habitat/controller/controller.go
```diff
@@ -145,20 +165,77 @@ func (hc *HabitatController) onAdd(obj interface{}) {
 						{
 							Name:  "habitat-service",
 							Image: sg.Spec.Image,
+							VolumeMounts: []apiv1.VolumeMount{
+								{
+									Name:      "config",
+									MountPath: "/hab/kubernetes-configmap",
```
LiliC: Can creating `/hab` directory cause any problems with habitat, as in does it interfere with habitat creating `/hab`.

pkg/habitat/controller/controller.go
```diff
@@ -145,20 +165,77 @@ func (hc *HabitatController) onAdd(obj interface{}) {
...
+							Name: "config",
+							VolumeSource: apiv1.VolumeSource{
+								ConfigMap: &apiv1.ConfigMapVolumeSource{
+									LocalObjectReference: apiv1.LocalObjectReference{
+										Name: configMapName(sg),
+									},
+									Items: []apiv1.KeyToPath{
+										{
+											Key:  "peer-watch-file",
+											Path: "peer-watch-file",
```
LiliC: I think `peer-ip` would be more descriptive what the file contains. Wdyt?

pkg/habitat/controller/controller.go
```diff
@@ -145,20 +165,77 @@ func (hc *HabitatController) onAdd(obj interface{}) {
...
+
+	// Create the ConfigMap for the peer watch file.
+	configMap := &apiv1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: configMapName(sg),
+			// Declare this ConfigMap to be owned by the Deployment, so that deleting
+			// the Deployment deletes the ConfigMap.
+			OwnerReferences: []metav1.OwnerReference{
+				{
+					APIVersion: "extensions/v1beta1",
```
LiliC: Lets be consistent and use `apps/....`, as we are using the `apps` api above in the deployment.

pkg/habitat/controller/controller.go
```diff
@@ -145,20 +165,78 @@ func (hc *HabitatController) onAdd(obj interface{}) {
 						{
 							Name:  "habitat-service",
 							Image: sg.Spec.Image,
+							VolumeMounts: []apiv1.VolumeMount{
+								{
+									Name:      "config",
+									MountPath: "/kubernetes-configmap",
```
LiliC: Maybe instead `/habitat-operator`, but feel free to ignore if you disagree.

pkg/habitat/controller/utils.go
asymmetric: Yes, it's unrelated, but I thought it's not worth creating a separate PR just for this.

pkg/habitat/controller/controller.go
asymmetric: Yeah, I don't really understand why the `Deployment` struct is defined in both places.

asymmetric: btw, `❯ kubectl get deploy -o yaml` returns

```bash

File: pkg/habitat/controller/controller.go [hunk 7]
```diff
@@ -145,20 +165,78 @@ func (hc *HabitatController) onAdd(obj interface{}) {
...
+			},
+		},
+		// Initially, the file will be empty. It will be populated by the
+		// controller once the first pod has gotten an IP assigned to it.
+		Data: map[string]string{
+			"peer-watch-file": "",
+		},
+	}
+
+	{
```
LiliC: I am curious, what is the purpose of the `{ ... }`.

asymmetric: We would have to name the `err` variable something specific otherwise. Because there is already a variable named `err` above, and `:=` always wants new variable names. Go stuff.

LiliC: Cool, TIL! Thanks. 
But you can just do `err = whatever()` and that will override the `err`.

asymmetric: RIGHT!

PR #29, opened 2017-07-26
Add support for --group flag

### LINKED ISSUE #23: Add support for  --group flag
The habitat client supports the `--group` flag, which allows users to start the supervisors in a specific group.

In order to support this, we need to:

- [ ] Add a `group` key to the CRD
- [ ] Pass the `--group` flag as arguments to to the containers

asymmetric:
The value of the `group` key in the CRD is passed to the command line flag `--group`.

If none is specified, we use the value "default", which is the same as
not specifying any --group.

Passing arguments like this works because the images have `init.sh` as their [`ENTRYPOINT`](https://github.com/habitat-sh/habitat/blob/a66961df540e0ced8acf8d94d0b6e9aa79d322b8/components/pkg-dockerize/bin/hab-pkg-dockerize.sh#L139).

`init.sh` simply [runs](https://github.com/habitat-sh/habitat/blob/a66961df540e0ced8acf8d94d0b6e9aa79d322b8/components/studio/libexec/hab-studio-type-baseimage.sh#L122)`hab sup start core/nginx --group default`

Closes #23.

### DIFF
--- a/examples/service_group-leader.yml
+++ b/examples/service_group-leader.yml
@@ -3,8 +3,11 @@ kind: ServiceGroup
 metadata:
   name: example-leader-follower-service-group
 spec:
-  topology: leader-follower
   # the core/consul habitat service packaged as a Docker image
   image: kinvolk/consul-hab
   # count must be at least 3 for a standalone topology
   count: 3
+  habitat:
+    topology: leader-follower
+    # if not present, defaults to "default"
+    group: foobar
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -138,8 +138,8 @@ func (hc *HabitatController) onAdd(obj interface{}) {
 	level.Debug(hc.logger).Log("msg", "validated object")
 
 	group := "default"
-	if sg.Spec.Group != "" {
-		group = sg.Spec.Group
+	if sg.Spec.Habitat.Group != "" {
+		group = sg.Spec.Habitat.Group
 	}
 
 	// This value needs to be passed as a *int32, so we convert it, assign it to a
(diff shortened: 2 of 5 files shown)


asymmetric:
Thinking more about this, I think there's a naming problem here: the CRD is called `ServiceGroup`, and the `name` key sets the Kubernetes Service Group's name. But we also have a key named `group`, to change the Habitat Service Group's name. 

We could solve this by changing the CRD name to something like `HabitatService`.

WDYT?

pkg/habitat/controller/controller.go
```diff
@@ -137,12 +137,18 @@ func (hc *HabitatController) onAdd(obj interface{}) {
 
 	level.Debug(hc.logger).Log("msg", "validated object")
 
-	// Create a deployment.
+	var group string
```
LiliC: Wdyt of doing something like:
```
group := "default"
if sg.Spec.Group != "" {
   group = sg.Spec.Group
}
```

Also is there any other validation we should be doing in the operator, or do we let habitat take care of validating group value?

asymmetric:
PTALA.

pkg/habitat/controller/controller.go
asymmetric: I would say in this case, we delegate validation to habitat. But I don't have a hard and fast rule to decide when we should do it.

examples/service_group-leader.yml
```diff
@@ -3,8 +3,11 @@ kind: ServiceGroup
 metadata:
   name: example-leader-follower-service-group
 spec:
-  topology: leader-follower
   # the core/consul habitat service packaged as a Docker image
   image: kinvolk/consul-hab
   # count must be at least 3 for a standalone topology
   count: 3
+  habitat:
+    topology: leader
```
LiliC: Did you mean `leader-follower` here?

asymmetric: The corresponding CLI flag is `-t leader`.

PR #30, opened 2017-07-27
Rename CRD from ServiceGroup to HabitatService

asymmetric:
This is to avoid ambiguity between the ServiceGroup CRD, and the service
group concept in Habitat, selected with the `--group` flag in the CLI.

See https://github.com/kinvolk/habitat-operator/pull/29

### DIFF
--- a/examples/crd.yml
+++ b/examples/crd.yml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   # name must match the spec fields below, and be in the form: <plural>.<group>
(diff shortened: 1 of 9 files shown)


asymmetric:
We decided not to rename the CR. Instead, we add a `habitat` section to it, that contains the flags passed to the `hab` tool:

```yaml
apiVersion: habitat.sh/v1
kind: ServiceGroup
metadata:
  name: example-standalone-service-group
spec:
  topology: standalone
  # the core/nginx habitat service packaged as a Docker image
  image: kinvolk/nginx-hab
  count: 1
```

PR #32, opened 2017-07-27
write pod IP in the ConfigMap

LiliC:
We watch for Pods and on update once one of the pods is running we get its IP and write to the ConfigMap. This pod has the label `habitat`.

This closes https://github.com/kinvolk/habitat-operator/issues/11.
In next couple of PRs I will tackle the `TODO`'s.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -27,11 +27,17 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
 
 	crv1 "github.com/kinvolk/habitat-operator/pkg/habitat/apis/cr/v1"
+)
+
+const (
+	resyncPeriod = 1 * time.Minute
+	peerFile     = "peer-watch-file"
 )
 
 type HabitatController struct {
@@ -67,15 +73,13 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
 	return hc, nil
 }
 
-// Run starts a Habitat resource controller
+// Run starts a Habitat resource controller.
 func (hc *HabitatController) Run(ctx context.Context) error {
 	level.Info(hc.logger).Log("msg", "Watching Service Group objects")
 
-	_, err := hc.watchCustomResources(ctx)
-	if err != nil {
-		level.Error(hc.logger).Log("msg", "Failed to register watch for ServiceGroup resource", "err", err)
-		return err
-	}
+	hc.watchCustomResources(ctx)
+
+	hc.watchPods(ctx)
 
 	// This channel is closed when the context is canceled or times out.
 	<-ctx.Done()
@@ -84,7 +88,7 @@ func (hc *HabitatController) Run(ctx context.Context) error {
 	return ctx.Err()
 }
 
-func (hc *HabitatController) watchCustomResources(ctx context.Context) (cache.Controller, error) {
+func (hc *HabitatController) watchCustomResources(ctx context.Context) {
 	source := cache.NewListWatchFromClient(
 		hc.config.HabitatClient,
 		crv1.ServiceGroupResourcePlural,
@@ -100,7 +104,7 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) (cache.Co
 		// resyncPeriod
 		// Every resyncPeriod, all resources in the cache will retrigger events.
 		// Set to 0 to disable the resync.
-		1*time.Minute,
+		resyncPeriod,
 
 		// Your custom resource event handlers.
 		cache.ResourceEventHandlerFuncs{
@@ -111,8 +115,6 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) (cache.Co
 
 	// The k8sController will start processing events from the API.
 	go k8sController.Run(ctx.Done())
-
-	return k8sController, nil
 }
 
 func (hc *HabitatController) onAdd(obj interface{}) {
@@ -152,7 +154,8 @@ func (hc *HabitatController) onAdd(obj interface{}) {
 			Template: apiv1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
-						"habitat": "true",
+						"habitat":       "true",
+						"service-group": sg.Name,
 					},
 				},
 				Spec: apiv1.PodSpec{
@@ -176,11 +179,11 @@ func (hc *HabitatController) onAdd(obj interface{}) {
 							VolumeSource: apiv1.VolumeSource{
 								ConfigMap: &apiv1.ConfigMapVolumeSource{
 									LocalObjectReference: apiv1.LocalObjectReference{
-										Name: configMapName(sg),
+										Name: configMapName(sg.Name),
 									},
 									Items: []apiv1.KeyToPath{
 										{
-											Key:  "peer-watch-file",
+											Key:  peerFile,
 											Path: "peer-ip",
 										},
 									},
@@ -202,27 +205,7 @@ func (hc *HabitatController) onAdd(obj interface{}) {
 	level.Info(hc.logger).Log("msg", "created deployment", "name", d.GetObjectMeta().GetName())
 
 	// Create the ConfigMap for the peer watch file.
-	configMap := &apiv1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: configMapName(sg),
-			// Declare this ConfigMap to be owned by the Deployment, so that deleting
-			// the Deployment deletes the ConfigMap.
-			OwnerReferences: []metav1.OwnerReference{
-				{
-					APIVersion: "extensions/v1beta1",
-					Kind:       "Deployment",
-					Name:       sg.Name,
-					UID:        d.UID,
-				},
-			},
-		},
-		// Initially, the file will be empty. It will be populated by the
-		// controller once the first pod has gotten an IP assigned to it.
-		Data: map[string]string{
-			"peer-watch-file": "",
-		},
-	}
-
+	configMap := newConfigMap(sg.Name, d.UID, "")
 	_, err = hc.config.KubernetesClientset.CoreV1Client.ConfigMaps(apiv1.NamespaceDefault).Create(configMap)
 	if err != nil {
 		level.Error(hc.logger).Log("msg", err)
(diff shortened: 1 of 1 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -266,6 +269,88 @@ func (hc *HabitatController) onDelete(obj interface{}) {
 	level.Info(hc.logger).Log("msg", "deleted deployment", "name", deploymentName)
 }
 
+func (hc *HabitatController) watchPods(ctx context.Context) cache.Controller {
```
asymmetric: I don't know if it makes sense for this functoin to return the `Controller`. Why do you think it does?

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +269,88 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+	for i := range listPods.Items {
+		p := &listPods.Items[i]
+		if p.Status.Phase == apiv1.PodRunning {
+			return p, nil
+		}
+	}
+	return nil, nil
+}
+
+func ListOptions() metav1.ListOptions {
```
asymmetric: Can't this just be a variable?

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +269,88 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+	return metav1.ListOptions{
+		LabelSelector: fields.SelectorFromSet(fields.Set(map[string]string{
+			"habitat": "true",
+		})).String(),
+	}
+}
+
+func (hc *HabitatController) writeIP(sg string, ip string) error {
+	cmname := fmt.Sprintf("%s-peer-file", sg)
```
asymmetric: Isn't there already the `configMapName` function for this?

Also, the var should be camelCased: `cmName`. Same for a few vars below.

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +269,88 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+	cm, err := hc.config.KubernetesClientset.CoreV1().ConfigMaps(apiv1.NamespaceDefault).Get(cmname, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	updatedcm := cm
```
asymmetric: We should make a deep copy here, i.e. copy each field of the CM individually. Otherwise they will share memory in the nested fields, and we are contravening [this](https://github.com/kubernetes/apiextensions-apiserver/blob/c682349b0d1c12975d8e24a9799b66747255d7a5/examples/client-go/controller/controller.go#L86-L93).

I'm pretty sure that standard k8s resources have some tooling already implemented for that, but otherwise we might have to do something [like this](https://github.com/kubernetes/apiextensions-apiserver/blob/c682349b0d1c12975d8e24a9799b66747255d7a5/examples/client-go/apis/cr/v1/zz_generated.deepcopy.go#L81).

pkg/habitat/controller/controller.go
LiliC: Only followed the `watchCustomResources`as an example to be consistent, but I agree will remove. Do you think we can remove from there as well?

LiliC: Also was not sure about returning an error, as I do not receive any errors, what do you think?

asymmetric: Yeah let's remove it from both, we can always readd.

asymmetric: Same for the error, we can safely remove it.

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +269,88 @@ func (hc *HabitatController) onDelete(obj interface{}) {
 	level.Info(hc.logger).Log("msg", "deleted deployment", "name", deploymentName)
 }
 
+func (hc *HabitatController) watchPods(ctx context.Context) cache.Controller {
+	cwl := cache.NewListWatchFromClient(
+		hc.config.KubernetesClientset.CoreV1().RESTClient(),
+		"pods",
+		apiv1.NamespaceAll,
+		fields.Everything())
```
asymmetric: Is this the place where we could say "only list/watch pods that have certain lables set"?

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
 	level.Info(hc.logger).Log("msg", "deleted deployment", "name", deploymentName)
 }
 
+func (hc *HabitatController) watchPods(ctx context.Context) {
+	cwl := cache.NewListWatchFromClient(
```
asymmetric: Think this shoud be `clw`.

pkg/habitat/controller/controller.go
LiliC: Yes exactly, I plan on doing that as part of the next PR, if thats ok with you?

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+		})
+
+	go c.Run(ctx.Done())
+}
+
+func (hc *HabitatController) onPodAdd(obj interface{}) {
+}
+
+func (hc *HabitatController) onPodUpdate(oldObj, newObj interface{}) {
+	// TODO: Do not retrive or write IP if we are deploying standalone topology
```
asymmetric: Retrieve

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+	if pod != nil {
+		err = hc.writeIP(pod.ObjectMeta.Labels["service-group"], pod.Status.PodIP)
+		if err != nil {
+			level.Error(hc.logger).Log("msg", err)
+		}
+	}
+}
+
+func (hc *HabitatController) onPodDelete(obj interface{}) {
+	// TODO: Make sure the pod that was deleted was not the same pod
```
asymmetric: Here and elsewhere: I was convinced some time ago to write comments like sentences, i.e. start with capital and end with a dot.

If you like that, please update the comments here, so we keep consistent.

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+
+	go c.Run(ctx.Done())
+}
+
+func (hc *HabitatController) onPodAdd(obj interface{}) {
+}
+
+func (hc *HabitatController) onPodUpdate(oldObj, newObj interface{}) {
+	// TODO: Do not retrive or write IP if we are deploying standalone topology
+	pod, err := hc.getPod()
```
asymmetric: Why do we need to get the pod again? We just got it, it's the `newObj`.

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+	}
+}
+
+func (hc *HabitatController) onPodDelete(obj interface{}) {
+	// TODO: Make sure the pod that was deleted was not the same pod
+	// whose IP was in the ConfigMap
+}
+
+func (hc *HabitatController) getPod() (*apiv1.Pod, error) {
+	listOptions := metav1.ListOptions{
```
asymmetric: I don't think we need this function. We are already listing as part of the `ListWatch`, and there is where we should define our label selectors.

When we get an event, that's our Pod, we cast it and do what we have to do. :)

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+	for i := range listPods.Items {
+		p := &listPods.Items[i]
+		if p.Status.Phase == apiv1.PodRunning {
+			return p, nil
+		}
+	}
+	return nil, nil
+}
+
+func (hc *HabitatController) writeIP(sg string, ip string) error {
```
asymmetric: Why does this function need the `sg` argument?

IIRC, @krnowak said he only needs `IP:PORT` in the file. In which scenario does it make sense to have the SG name too?

LiliC: @asymmetric We need the sg to get the ConfigMap name, that happens below this comment.

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+	updatedcm := cm
+	updatedcm.Data["peer-watch-file"] = ip
+	_, err = hc.config.KubernetesClientset.CoreV1().ConfigMaps(apiv1.NamespaceDefault).Update(updatedcm)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func configMapName(sgName string) string {
```
asymmetric: OK, this is pretty unimportant, but I liked it more when this function took the whole struct as an argument, for two reasons:

a) it's more generic, we might in the future not use only the name
b) it keeps the knowledge of how to extract a name embedded within the function. So better encapsulation.

But yeah, not so important :)

pkg/habitat/controller/controller.go
```diff
@@ -67,15 +67,13 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
...
+// Run starts a Habitat resource controller.
 func (hc *HabitatController) Run(ctx context.Context) error {
 	level.Info(hc.logger).Log("msg", "Watching Service Group objects")
 
-	_, err := hc.watchCustomResources(ctx)
-	if err != nil {
-		level.Error(hc.logger).Log("msg", "Failed to register watch for ServiceGroup resource", "err", err)
-		return err
-	}
+	hc.watchCustomResources(ctx)
```
asymmetric: I wonder if it wouldn't be more clear if we had `go hc.watchCustomResources(ctx)` here. So we would have to change that function and remove the `go` there.

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+	clw := cache.NewListWatchFromClient(
+		hc.config.KubernetesClientset.CoreV1().RESTClient(),
+		"pods",
+		apiv1.NamespaceAll,
+		fields.Everything())
+
+	_, c := cache.NewInformer(
+		clw,
+		&apiv1.Pod{},
+		60*time.Second,
```
asymmetric: We should probably be consistent here and use `1 * time.Minute` like above. Or `Second` everywhere :)

asymmetric: We could also make this into a constant at the top of the file.

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+func (hc *HabitatController) onPodAdd(obj interface{}) {
+}
+
+func (hc *HabitatController) onPodUpdate(oldObj, newObj interface{}) {
+	// TODO: Do not retrieve or write IP if we are deploying a standalone topology.
+	pod, err := hc.getPod()
+	if err != nil {
+		level.Error(hc.logger).Log("msg", err)
+	}
+	if pod != nil {
```
asymmetric: I don't think you need to perform this check in this case, but as a general point: I think it's better (and more idiomatic) to deal with the exceptional case inside the `if` block, and in this case `return`, and have the "happy path" have no indentation.

[Here](https://golang.org/doc/effective_go.html#if) is some info about this:
>  The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise. Since error cases tend to end in return statements, the resulting code needs no else statements.

pkg/habitat/controller/controller.go
```diff
@@ -266,6 +263,83 @@ func (hc *HabitatController) onDelete(obj interface{}) {
...
+
+func (hc *HabitatController) writeIP(sg string, ip string) error {
+	cmName := configMapName(sg)
+	cm, err := hc.config.KubernetesClientset.CoreV1().ConfigMaps(apiv1.NamespaceDefault).Get(cmName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	updatedcm := cm
+	updatedcm.Data["peer-watch-file"] = ip
```
asymmetric: I think it's time we turn `peer-watch-file` into a constant :)

pkg/habitat/controller/controller.go
LiliC: I agree, but right now we only need the `sgName`, so I would say we don't need to make it more generic for now. :) If thats ok with you?

asymmetric: Haha yes, it's OK.

pkg/habitat/controller/controller.go
asymmetric: I think it should be part of this one, it's central to what this PR is about IMO.

pkg/habitat/controller/controller.go
```diff
@@ -268,7 +268,7 @@ func (hc *HabitatController) watchPods(ctx context.Context) {
 		hc.config.KubernetesClientset.CoreV1().RESTClient(),
 		"pods",
 		apiv1.NamespaceAll,
-		fields.Everything())
+		fields.SelectorFromSet(fields.Set(map[string]string{"habitat": "true"})))
```
asymmetric: I think we could do something like `fields.Matches(label)`

pkg/habitat/controller/controller.go
```diff
@@ -288,42 +288,25 @@ func (hc *HabitatController) onPodAdd(obj interface{}) {
 
 func (hc *HabitatController) onPodUpdate(oldObj, newObj interface{}) {
 	// TODO: Do not retrieve or write IP if we are deploying a standalone topology.
-	pod, err := hc.getPod()
+	pod := newObj.(*apiv1.Pod)
```
asymmetric: Here you can do it as we do above, i.e. `pod, ok := ...` and then check `ok`.

asymmetric: If `!ok`, it's an error, not just a `return`, same as above.

pkg/habitat/controller/controller.go
```diff
@@ -288,42 +288,25 @@ func (hc *HabitatController) onPodAdd(obj interface{}) {
...
+	if pod == nil {
+		return
+	}
+	if pod.Status.Phase != apiv1.PodRunning {
```
asymmetric: Could this be part of the filters too, so that we only get events for running pods?

pkg/habitat/controller/controller.go
```diff
@@ -288,42 +288,25 @@ func (hc *HabitatController) onPodAdd(obj interface{}) {
...
+		return
+	}
+
+	err := hc.writeIP(pod.ObjectMeta.Labels["service-group"], pod.Status.PodIP)
```
asymmetric: Here, I kind of think it would be better to just pass the pod, and let the method figure out what/how to do it. See previous comment about encapsulation.

pkg/habitat/controller/controller.go
asymmetric: Please disregard this. We would likely introduce race conditions.

PR #33, opened 2017-07-28
*: automate building operator

LiliC:
Add make build and update .gitignore file.

Makefile will also be useful later on in the future as we will likely extend this to build binaries and container images. For now it just makes life a bit easier. :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,4 +13,5 @@
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
 .glide/
 
-cmd/operator/operator
+/operator
+
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,4 @@
+build:
+	go build -i github.com/kinvolk/habitat-operator/cmd/operator
+
+.PHONY: build


.gitignore
```diff
@@ -13,4 +13,4 @@
 # Project-local glide cache, RE: https://github.com/Masterminds/glide/issues/736
 .glide/
 
-cmd/operator/operator
+operator
```
krnowak: I'd prefer to have `/operator`, so we don't accidentally ignore all the `operator` files/directories in the directory tree.

Makefile
```diff
@@ -0,0 +1,2 @@
+build:
+	go build -i github.com/kinvolk/habitat-operator/cmd/operator
```
krnowak: Add `.PHONY: build` below, so if you create a `build` file, then make will know to ignore it and run the build rule.

krnowak: What this will do anyway? Will it download the project and build it instead of building the sources from this directory?

LiliC: No, it expects it to be in your GOPATH and it builds from there.

LiliC:
@krnowak Can you have another look, thanks.

krnowak:
LFAD.

Issue #22, opened 2017-07-25
Update IP in ConfigMap when pod dies


LiliC:
In the case when we already have the pod IP data in the ConfigMap, but the pod with that IP has died we need to select a new pod and update the ConfigMap with that IP.

For more context see https://github.com/kinvolk/habitat-operator/issues/21 and https://github.com/kinvolk/habitat-operator/issues/11

PR #35, opened 2017-08-01
Pod IP file: Handle edge cases and pod deletion 

LiliC:
This PR handles updating ConfigMap IP file on pod deletion with a new IP from a running pod. There were also some edge cases I found that are now also handled. 

It also introduces a modified  newListWatchFromClient so we are able to filter `habitat` label and that way only get pods marked with that label, as we only care about watching those pods. The existing function can only filter by fields selector.

Closes https://github.com/kinvolk/habitat-operator/issues/22.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -310,7 +310,11 @@ func (hc *HabitatController) onPodDelete(obj interface{}) {
 	if pod == nil {
 		return
 	}
-	sgName := pod.ObjectMeta.Labels["service-group"]
+	sgName, exists := pod.ObjectMeta.Labels["service-group"]
+	if !exists {
+		level.Error(hc.logger).Log("msg", "Could not retrieve service group name because label did not exist.")
+		return
+	}
 	cmName := configMapName(sgName)
 	cm, err := hc.config.KubernetesClientset.CoreV1().ConfigMaps(apiv1.NamespaceDefault).Get(cmName, metav1.GetOptions{})
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -286,38 +288,89 @@ func (hc *HabitatController) onPodUpdate(oldObj, newObj interface{}) {
...
+	if oldIP != "" {
+		// Do not overwrite IP with itself.
+		if ip == oldIP {
+			return nil
+		}
+		podList, err := hc.config.KubernetesClientset.CoreV1().Pods(apiv1.NamespaceDefault).List(metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+		for i := range podList.Items {
```
krnowak: You probably could change it to `for _, oldPod := range podlist.Items`, no?

pkg/habitat/controller/controller.go
```diff
@@ -286,38 +288,89 @@ func (hc *HabitatController) onPodUpdate(oldObj, newObj interface{}) {
...
-	// whose IP was in the ConfigMap.
+	pod, ok := obj.(*apiv1.Pod)
+	if !ok {
+		level.Error(hc.logger).Log("msg", "Failed to cast pod.")
+		return
+	}
+	if pod == nil {
+		return
+	}
+	sgName := pod.ObjectMeta.Labels["service-group"]
```
krnowak: Can we be sure that this field always exist?

pkg/habitat/controller/controller.go
```diff
@@ -286,38 +288,89 @@ func (hc *HabitatController) onPodUpdate(oldObj, newObj interface{}) {
...
+	// Get only those pods that are running.
+	fs := fields.SelectorFromSet(fields.Set{
+		"status.phase": "Running",
+	})
+	podList, err := hc.config.KubernetesClientset.CoreV1().Pods(apiv1.NamespaceDefault).List(metav1.ListOptions{FieldSelector: fs.String()})
+	if err != nil {
+		level.Error(hc.logger).Log("msg", err)
+		return
+	}
+	for i := range podList.Items {
```
krnowak: How about `for _, newPod := range podList.Items`?

krnowak:
LFAD. Feel free to merge when your testing confirms that your changes work.

PR #40, opened 2017-08-10
Initial Configurations

LiliC:
This PR introduces initial configurations.

user.toml file serves as a way of initial configuration in Habitat. Our operator mounts this file in a Volume as a Secret that a user has to create first. (As examples in the examples/nodejs)  Habitat will pick up this file and all the configurations in this file will override the ones defined in default.toml. Changing this secret will not be picked up by our service. This work will be don in the next part, as part of configuration updates.

Was not sure how to name the field, I went with `toml`, but am open for better suggestions.


examples/nodejs/service_group-nodejs.yml
```diff
@@ -1,10 +1,42 @@
+apiVersion: v1
```
asymmetric: Do we need `-nodejs` in the filename? It's already in the path.

examples/nodejs/service_group-nodejs.yml
```diff
@@ -1,10 +1,42 @@
...
+kind: Secret
+metadata:
+  name: user-toml-secret
+type: Opaque
+data:
+  # Each item needs to be encoded in base64, as Kuberentes
```
asymmetric: *Kubernetes

examples/nodejs/service_group-nodejs.yml
```diff
@@ -1,10 +1,42 @@
...
+  # expects that encoding.
+  # Plain text content: "message = 'Hello from our Habitat-Operator!'"
+  user.toml: bWVzc2FnZSA9ICdIZWxsbyBmcm9tIG91ciBIYWJpdGF0LU9wZXJhdG9yISc=
+---
 apiVersion: habitat.sh/v1
 kind: ServiceGroup
 metadata:
-  name: example-nodejs-service-group
+  # Name must be the same as your service name in Habitat.
```
asymmetric: Why? Could we add an explanation here, or is there a better place?

examples/nodejs/service_group-nodejs.yml
```diff
@@ -1,10 +1,42 @@
...
+    # Create Secret with the initial configuration you want.
+    toml: user-toml-secret
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: mytutorialapp
+spec:
+  selector:
+    service-group: mytutorialapp
```
asymmetric: It seems like using camelCase here (`serviceGroup`) would be more along the standard.

examples/service_group-leader.yml
```diff
@@ -5,7 +5,7 @@ metadata:
 spec:
```
asymmetric: Since we have sub-directories for projects, should we move this to something like `examples/consul`, or `example/leader`?

pkg/habitat/apis/cr/v1/types.go
```diff
@@ -48,6 +48,10 @@ type Habitat struct {
 	Group string `json:"group"`
 	// Topology is the value of the --topology flag for the hab client.
 	Topology `json:"topology"`
+	// Toml is the name of the Secret that user previously created.
```
asymmetric: the user

pkg/habitat/apis/cr/v1/types.go
```diff
@@ -48,6 +48,10 @@ type Habitat struct {
...
+	// The file with this name is mounted inside of the pod, habitat will
+	// use it for initial configuration of the service.
+	Toml string `json:"toml"`
```
asymmetric: What about `Config`, or `UserConfig`?

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,130 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+	updatedCM := newConfigMap(sgName, cm.UID, peerFile, ip)
 	_, err = hc.config.KubernetesClientset.CoreV1().ConfigMaps(apiv1.NamespaceDefault).Update(updatedCM)
 	if err != nil {
 		return err
 	}
 	return nil
 }
 
-func newConfigMap(sgName string, uid types.UID, ip string) *apiv1.ConfigMap {
+func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) *appsv1beta1.Deployment {
```
asymmetric: Does this need to be a method, or can it just be a function?

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,130 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+	// This value needs to be passed as a *int32, so we convert it, assign it to a
+	// variable and afterwards pass a pointer to it.
+	count := int32(sg.Spec.Count)
+
+	// Set the service arguments we send to Habitat.
+	habArgs := []string{}
```
asymmetric: You can just do `var habArgs []string` here.

examples/nodejs/service_group-nodejs.yml
LiliC: We use that name to generate the path where `user.toml` should be placed. Went for this idea for now, as I didn't want yet another field name, but I am open to suggestions.

Wasn't sure our user needs to know why we need this, but can add the explanation.

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,130 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+		// Lets make sure our secret is there before mounting it.
+		secret, err := hc.config.KubernetesClientset.CoreV1().Secrets(apiv1.NamespaceDefault).Get(sg.Spec.Habitat.Toml, metav1.GetOptions{})
+		if err != nil {
+			level.Error(hc.logger).Log("msg", err)
+			return base
+		}
+
+		if secret == nil {
+			// If we can't find the secret we should just return base deployment.
+			level.Error(hc.logger).Log("msg", "Secret could not be mounted as it did not exist.")
```
asymmetric: Is it really an error if we can't find the secret? Or is it just a warning?

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,130 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+			return base
+		}
+
+		secretVolumeMount := &apiv1.VolumeMount{
+			Name:      "initialconfig",
+			MountPath: mp,
```
asymmetric: I wonder if we even need the `mp` variable, since it seems to be used only here.

pkg/habitat/apis/cr/v1/types.go
LiliC: I think now that we are going in the direction of using the `user.toml` as a way to do config updates, `Config` would be better.

LiliC: On the other hand should our name hint to the fact this should be in the `.toml` format?

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,130 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+					},
+				},
+			},
+		},
+	}
+
+	// If we have a secret name present we should mount that secret.
+	if sg.Spec.Habitat.Toml != "" {
+		// Our user.toml file must be in the directory which name is service specific.
+		mp := fmt.Sprintf("/hab/svc/%s", sg.Name)
```
LiliC: `mount path`, will move. Are you ok with the naming?

asymmetric: Also, do we need to use this path? Or can we re-use the `/habitat-operator` where we put the ConfigMap?

LiliC: We need it, it's where [habitat](https://github.com/habitat-sh/habitat/blob/master/components/sup/src/manager/service/config.rs#L147) expects to find the file.

pkg/habitat/controller/controller.go
```diff
@@ -395,11 +445,11 @@ func newConfigMap(sgName string, uid types.UID, ip string) *apiv1.ConfigMap {
...
 		Data: map[string]string{
-			peerFile: ip,
+			key: value,
 		},
 	}
 }
 
-func configMapName(sgName string) string {
-	return fmt.Sprintf("%s-peer-file", sgName)
+func configMapName(sgName string, key string) string {
```
asymmetric: You can do `sgName, key string` here, since both have the same type.

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,130 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+		}
+
+		base.Spec.Template.Spec.Containers[0].VolumeMounts = append(base.Spec.Template.Spec.Containers[0].VolumeMounts, *secretVolumeMount)
+		base.Spec.Template.Spec.Volumes = append(base.Spec.Template.Spec.Volumes, *secretVolume)
+	}
+
+	return base
+}
+
+func newConfigMap(sgName string, uid types.UID, key string, value string) *apiv1.ConfigMap {
```
asymmetric: I would do `key, value string` here.

pkg/habitat/controller/controller.go
asymmetric: BTW, do we need the `key` in the name? Couldn't we just create a ConfigMap with name equal to `sgName`?

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,130 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+			return base
+		}
+
+		if secret == nil {
+			// If we can't find the secret we should just return base deployment.
+			level.Error(hc.logger).Log("msg", "Secret could not be mounted as it did not exist.")
+			return base
+		}
+
+		secretVolumeMount := &apiv1.VolumeMount{
```
asymmetric: Does it need to be a Secret because it can contain sensitive data?

pkg/habitat/controller/controller.go
LiliC: At first I was creating a CM instead of a Secret to store `user.toml`, which is why I added the `key`, to differentiate them on creation. But I left it, as I felt like it was a good way to describe what is in this CM. Can remove if you don't agree?

pkg/habitat/controller/controller.go
LiliC: Yes exactly, this will used to store configuration data and that can be anything including passwords.

pkg/habitat/controller/controller.go
asymmetric: Is this controlled by the `--config-from` path? Or is that flag for the `default.toml`?

pkg/habitat/controller/controller.go
asymmetric: Do we need to create many different CMs? I think one per SG would be enough, and we can just add keys to it as needed.

In that case, the `sgName` is enough for the CM.

pkg/habitat/controller/controller.go
LiliC: `user.toml` is used on initial configuration and is always the same path. This already exists in Habitat, but is not documented well. Order of precedence of which config is taken:
- env variables
- `user.toml`
- `default.toml`

`--config-from` is for specifying a custom directory from where you want to load your `plan.sh` and `default.toml` from what I [understand](https://www.habitat.sh/docs/reference/plan-syntax/).

pkg/habitat/apis/cr/v1/types.go
asymmetric: I don't think so. I think that's an implementation detail that we can mention in the README (and I'm not even sure we need to, since it's a requirement of Habitat, not the operator).

examples/nodejs/service_group-nodejs.yml
```diff
@@ -1,10 +1,42 @@
...
+  # Name must be the same as your service name in Habitat.
+  name: mytutorialapp
 spec:
   image: kinvolk/nodejs-hab
   count: 1
   habitat:
     topology: standalone
     group: nodejs
+    # Create Secret with the initial configuration you want.
+    toml: user-toml-secret
```
asymmetric: I wonder: shouldn't we (the operator) take care of turning the user-provided TOML config into a Secret?

Basically, in this YAML the user would embed a chunk of TOML, we would parse it, encode it and create the secret for them.

Less error-prone and cumbersome for the user.

WDYT?

pkg/habitat/apis/cr/v1/types.go
LiliC: `Config` is fine with me.

pkg/habitat/controller/controller.go
LiliC: Agreed.

examples/nodejs/service_group-nodejs.yml
LiliC: Yes, I was doing this in the start as well, had a field where user can place their `toml` chunk. But changed my mind since if this is a large file that has 100+ lines, passing it in a file is a better option, which is what you can do with `Secrets`. Or at least letting the user have that option of passing in a file.

And also `Secrets` from what I understand are/(can be) [encrypted](https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/#verifying-that-data-is-encrypted) in etcd, not sure about `CRD`s.

pkg/habitat/controller/controller.go
LiliC: Agreed it's not an error, good catch! Wdyt of `info`, as we don't really user `warning` so far? Also found [this](https://dave.cheney.net/2015/11/05/lets-talk-about-logging) interesting.

pkg/habitat/controller/controller.go
LiliC: Will drop key then. 👍

LiliC: Or wait until your [PR](https://github.com/kinvolk/habitat-operator/pull/47#pullrequestreview-56285223) is merged and rebase.

pkg/habitat/controller/controller.go
LiliC: Good suggestion. One thing I don't know how to handle, should we check for the Secret:
https://github.com/kinvolk/habitat-operator/blob/d728cc0865c38023a84eb0ed3801e90832271425/pkg/habitat/controller/controller.go#L391
in `onAdd` then? Or do we pass in our `KubernetesClientset`?

examples/service_group-leader.yml
LiliC: Sure can do that. Specially since you can just `kubectl` `create` or `apply` from a directory.

examples/nodejs/service_group-nodejs.yml
LiliC: This is just our label selector, it comes from [this](https://github.com/kinvolk/habitat-operator/blob/master/pkg/habitat/controller/controller.go#L163).

For that I followed the naming conventions from [here](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set). Changing the example to camelCase will not work here, but we can change the name of the label in a different PR if you want to?

pkg/habitat/controller/controller.go
asymmetric: I think it's a bit more than `Info`, so `Warning` makes more sense to me, but no strong preference.

pkg/habitat/controller/controller.go
```diff
@@ -387,25 +387,23 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) *appsv1beta1.D
...
+			// If we can't find the secret we should just return the base deployment.
 			level.Error(hc.logger).Log("msg", "Secret could not be mounted as it did not exist.")
 			return base
 		}
 
 		secretVolumeMount := &apiv1.VolumeMount{
-			Name:      "initialconfig",
-			MountPath: mp,
+			Name: "initialconfig",
+			// Our user.toml file must be in a directory with the same name as the service.
```
asymmetric: Service or Service Group? ;)

LiliC: I guess both should be the same? I am not sure its the correct approach. Let me know what you think [here](https://github.com/kinvolk/habitat-operator/pull/40#discussion_r132912376).

pkg/habitat/controller/controller.go
LiliC: Done.

examples/nodejs/service_group-nodejs.yml
asymmetric: Yes, I was suggesting changing the way this field is serialized to YAML across the board.

Let's do it in another PR.

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,128 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) *appsv1beta1.Deployment {
+	// This value needs to be passed as a *int32, so we convert it, assign it to a
+	// variable and afterwards pass a pointer to it.
+	count := int32(sg.Spec.Count)
+
+	// Set the service arguments we send to Habitat.
+	var habArgs []string
+	if sg.Spec.Habitat.Group != "" {
+		// When a service is started without explicitly naming the group,
+		// it's assigned to the default group for the name of that package.
```
asymmetric: I don't really understand the "for the name of that package" part :)

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,128 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+		// Let's make sure our secret is there before mounting it.
+		secret, err := hc.config.KubernetesClientset.CoreV1().Secrets(apiv1.NamespaceDefault).Get(sg.Spec.Habitat.Config, metav1.GetOptions{})
+		if err != nil {
+			level.Error(hc.logger).Log("msg", err)
+			return base
+		}
+
+		if secret == nil {
+			// If we can't find the secret we should just return the base deployment.
+			level.Warn(hc.logger).Log("msg", "Secret could not be mounted as it did not exist.")
```
asymmetric: Two things:

* Error messages [should not be capitalized](https://github.com/golang/go/wiki/Errors).
* I actually think this is an error, and we should just fail. Ignoring that the error couldn't be found is probably not what the user expects (see above). WDYT?

pkg/habitat/controller/controller.go
LiliC: I just took what it says in the [docs](https://www.habitat.sh/docs/concepts-services/). Agree, it's confusing, feel free to suggest better wording.

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,128 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+		},
+	}
+
+	// If we have a secret name present we should mount that secret.
+	if sg.Spec.Habitat.Config != "" {
+		// Let's make sure our secret is there before mounting it.
+		secret, err := hc.config.KubernetesClientset.CoreV1().Secrets(apiv1.NamespaceDefault).Get(sg.Spec.Habitat.Config, metav1.GetOptions{})
+		if err != nil {
+			level.Error(hc.logger).Log("msg", err)
+			return base
```
asymmetric: I think here we should return `nil, err`. I don't think we should recover and provide the `base` here: the user wanted a configured service, and if loading the config fails, the service will probably not be very useful anyway.

pkg/habitat/controller/controller.go
asymmetric: I'd just drop the last part, and leave: "When a service is started without explicitly naming the group, it's assigned to the default group"

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,128 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+		}
+
+		if secret == nil {
+			// If we can't find the secret we should just return the base deployment.
+			level.Warn(hc.logger).Log("msg", "Secret could not be mounted as it did not exist.")
+			return base
+		}
+
+		secretVolumeMount := &apiv1.VolumeMount{
```
asymmetric: I'd swap the order of the mount and the volume, just cause it makes more sense logically.

I know we don't do that for the ConfigMap mount above, but that's cause I thought having the `Containers` section first made more sense :)

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,128 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+			Name: "initialconfig",
```
asymmetric: We could turn this into a constant, since it's used in more than one place already.

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,128 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+			// Our user.toml file must be in a directory with the same name as the service.
+			MountPath: fmt.Sprintf("/hab/svc/%s", sg.Name),
+			ReadOnly:  false,
```
asymmetric: Why not make it read-only? Can this be changed during the lifetime of the supervisor, e.g. via gossip?

pkg/habitat/controller/controller.go
```diff
@@ -373,18 +311,128 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+						{
+							Key:  userTomlFile,
+							Path: userTomlFile,
+						},
+					},
+				},
+			},
+		}
+
+		base.Spec.Template.Spec.Containers[0].VolumeMounts = append(base.Spec.Template.Spec.Containers[0].VolumeMounts, *secretVolumeMount)
```
asymmetric: I'd also swap the order here.

pkg/habitat/controller/controller.go
LiliC: I only changed from error to warning because of this [comment](https://github.com/kinvolk/habitat-operator/pull/40#discussion_r132928690).

asymmetric:
A general point that I think got somehow lost in the mess of "outdated diffs":

We could alleviate the job of the user by creating the secret ourselves. The user would just have to embed the TOML in the `Config` field of the ServiceGroup.

We can also do this in another PR.

examples/nodejs/service_group-nodejs.yml
asymmetric: Good point about encryption. Let's leave it like this then.

examples/nodejs/service_group-nodejs.yml
asymmetric: OK, I think it's fine not to have an explanation here then.

pkg/habitat/controller/controller.go
asymmetric: Ah, I hadn't seen that we use `hc` there. I guess then it can't be a function.

Or otherwise, it can just be brought back into the parent method.

pkg/habitat/controller/controller.go
asymmetric: What I mean is that we should not return `base`, but instead `nil, err`.

asymmetric: And btw I was wrong about capitalization:

> This does not apply to logging, which is implicitly line-oriented and not combined inside other messages.

LiliC: Ah, yes makes perfect sense. Maybe returning an error msg like `"Secret %s not found for ServiceGroup %s"` wdyt? Should we keep the log as well, or not?

pkg/habitat/controller/controller.go
LiliC: Good point. What do you think of the name? As we are going into direction of this eventually serving as a source of all configuration, not sure `initialconfig` is a good name anymore.

We already have [config](https://github.com/kinvolk/habitat-operator/blob/master/pkg/habitat/controller/controller.go#L176). Any suggestions?

pkg/habitat/controller/controller.go
LiliC: This directory is also used by Habitat itself to store an array of different files and dir.

pkg/habitat/controller/controller.go
asymmetric: I'm not sure. Could be that the error message that's already there is enough. I'd try that first, i.e. returning `nil, err`. We can change it later.

pkg/habitat/controller/controller.go
LiliC: Done.

pkg/habitat/controller/controller.go
LiliC: Done.

pkg/habitat/controller/controller.go
LiliC: Done.

asymmetric:
Just needs to be rebased and then LEGIT TO ME.

pkg/habitat/controller/controller.go
asymmetric: Should we just move everything (i.e. for now, the leader IP) to the `user.toml`?

On the one hand, I see less and less need for the `ConfigMap` as a place for configuration, as most everything will actually be in the `user.toml`.

On the other hand, if we use the `user.toml` for the peer IP, that makes watching it much harder, because it might change for completely unrelated reasons.

WDYT? /cc @krnowak

LiliC: Not sure, IMHO I think they should be separate as its two "configuration" concepts, configuration updates for the Service and IP info is for the ring itself. I don't know, it's up to @krnowak to decide.

krnowak: The contents of the user.toml are provided by the user, the contents of the peer ip file is not.

PR #43, opened 2017-08-11
fix ConfigMaps OwnReference on update

LiliC:
When updating our CM with the new IP we were passing in the wrong UID.
Instead of passing in the UID of our deployment we were passing in
the UID of ourselves to serve as a Owner Reference. What this caused was
that our CM was not destroyed when deleting our deployment.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -373,7 +373,14 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
 			}
 		}
 	}
-	updatedCM := newConfigMap(sgName, cm.UID, ip)
+
+	// We need to retrieve our deployment to get the UID for the OwnerReference.
+	d, err := hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(apiv1.NamespaceDefault).Get(sgName, metav1.GetOptions{})
+	if err != nil {
+		return err
+	}
+
+	updatedCM := newConfigMap(sgName, d.UID, ip)
 	_, err = hc.config.KubernetesClientset.CoreV1().ConfigMaps(apiv1.NamespaceDefault).Update(updatedCM)
 	if err != nil {
 		return err
@@ -381,7 +388,7 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
 	return nil
 }
 
-func newConfigMap(sgName string, uid types.UID, ip string) *apiv1.ConfigMap {
+func newConfigMap(sgName string, parentUID types.UID, ip string) *apiv1.ConfigMap {
 	return &apiv1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: configMapName(sgName),
@@ -390,7 +397,7 @@ func newConfigMap(sgName string, uid types.UID, ip string) *apiv1.ConfigMap {
 					APIVersion: "extensions/v1beta1",
 					Kind:       "Deployment",
 					Name:       sgName,
-					UID:        uid,
+					UID:        parentUID,
 				},
 			},
 		},


pkg/habitat/controller/controller.go
```diff
@@ -373,7 +373,23 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+			"service-group": sgName,
+		}).String(),
+	}
+	d, err := hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(apiv1.NamespaceDefault).List(o)
+	if err != nil {
+		return err
+	}
+	if len(d.Items) != 1 {
+		// We only will ever have 1 deployment for each ConfigMap.
+		return fmt.Errorf("We received %d deployment(s) with label 'service-group': %s, but were expecting 1.", len(d.Items), sgName)
```
asymmetric: I've used `errors.New` [in the past](https://github.com/kinvolk/habitat-operator/blob/master/pkg/habitat/controller/controller.go#L57). Which one do you think is more idiomatic?

We should settle on one.

LiliC: I believe both make sense, as its different scenarios yes.

pkg/habitat/controller/controller.go
```diff
@@ -373,7 +373,23 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
 	}
-	updatedCM := newConfigMap(sgName, cm.UID, ip)
+
+	// We need to retrieve our deployment to get the UID for the OwnerReference.
+	o := metav1.ListOptions{
+		LabelSelector: labels.SelectorFromSet(map[string]string{
+			"service-group": sgName,
+		}).String(),
+	}
+	d, err := hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(apiv1.NamespaceDefault).List(o)
```
asymmetric: Can't we use `Get` and pass it the Deployment name? This way we also don't have to check for > 1, since names must be unique.

pkg/habitat/controller/controller.go
```diff
@@ -373,7 +373,23 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+	if err != nil {
+		return err
+	}
+	if len(d.Items) != 1 {
+		// We only will ever have 1 deployment for each ConfigMap.
+		return fmt.Errorf("We received %d deployment(s) with label 'service-group': %s, but were expecting 1.", len(d.Items), sgName)
+	}
+
+	updatedCM := newConfigMap(sgName, d.Items[0].UID, ip)
```
asymmetric: I think it would help to add documentation to the `newConfigMap` function, to explain what the fields represent.

Or maybe, just changing the name of the `uid` param to `parentUID`.

pkg/habitat/controller/controller.go
```diff
@@ -373,7 +373,23 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
 			}
 		}
 	}
-	updatedCM := newConfigMap(sgName, cm.UID, ip)
+
+	// We need to retrieve our deployment to get the UID for the OwnerReference.
```
asymmetric: Doesn't the existing CM already have the correct UID?

LiliC: We don't know that. What we do know is that we want our CM to have the deployment `UID` as an `OwnerReference`, so we pass that instead of assuming/retrieving it from our CM. Using and mutating existing state can cause problems, this way we are always sure we have the correct state afterwards.

And if we don't want to always do a request to the kubernetes API  then we can introduce a cache for the deployments as well.

LiliC:
Can you have another look @asymmetric. The error msg you are getting happens in master as well, its separate issue from this PR.

pkg/habitat/controller/controller.go
asymmetric: I think we do know that, since we assign the UID when we first create the CM. Here we retrieve an already existing one, so the UID is already set.

Also, I'm not talking about modifying anything, just reading from the existing CM.

Or am I missing something?

asymmetric:
I don't get the error in master :confused:

pkg/habitat/controller/controller.go
LiliC: My thought process behind this was yes, we do set it, but we don't know if something/someone modifies it. If we can retrieve it from the source of truth, our deployment, then we are sure its the correct `OwnerReference` and when we destroy our CM we are certain this the correct one.

But if you insist that getting it from CM is better, will do that.

LiliC:
Huh? When I delete my CM and deployment (both separately, as `OwnerReference` doesn't work hence this PR), I get the following error in master:
```
level=error component=controller msg="configmaps \"example-leader-follower-service-group-peer-file\" not found"
```

asymmetric:
Ah I see, I was only deleting the Deployment and expecting to see an error. Ignore me :)

LiliC:
No problem. :) But I agree we need to fix that, already have it noted down, just not sure it has anything to do with this PR.

pkg/habitat/controller/controller.go
asymmetric: I think we can safely assume that no-one will change this apart from us, so I wouldn't worry about getting it straight from the Deployment.

I also think we could simplify this a bit by making a deep copy, changing the IP, and updating the CM. But I can do this in another PR if you don't feel like doing it. :)

PR #45, opened 2017-08-14
Badge and other small fixes

LiliC:
Badges were not linking to Travis and Go report card, but rather to the images.

Also included are some issues from [Go report card](https://goreportcard.com/report/github.com/kinvolk/habitat-operator).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-![travis] ![report]
+[![Build Status](https://travis-ci.org/kinvolk/habitat-operator.svg?branch=master)](https://travis-ci.org/kinvolk/habitat-operator) 
+[![Go Report Card](https://goreportcard.com/badge/github.com/kinvolk/habitat-operator)](https://goreportcard.com/report/github.com/kinvolk/habitat-operator)
 
 # habitat-operator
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,4 +1,5 @@
-![travis] ![report]
+[![Build Status](https://travis-ci.org/kinvolk/habitat-operator.svg?branch=master)](https://travis-ci.org/kinvolk/habitat-operator)
```
asymmetric: I would leave all links at the bottom. Alternatively, we should inline the other link (`crd`) as well.

pkg/habitat/apis/cr/v1/types.go
```diff
@@ -21,7 +21,7 @@ import (
 const ServiceGroupResourcePlural = "servicegroups"
 
 type ServiceGroup struct {
-	metav1.TypeMeta   `json:,inline"`
+	metav1.TypeMeta   `json:",inline"`
```
asymmetric: Good catch.

README.md
asymmetric: I see why it's complicated to move these to the bottom. There are 2 links being created here.

Feel free to ignore this comment (i.e., these 2 image links are inlined, all the rest at the bottom)

LiliC: 👍 Cool. Anything else or can we merge this one then? :)

PR #46, opened 2017-08-14
Deep copy ConfigMap instead of creating anew

asymmetric:
We use the provided machinery to create a deep copy of the ConfigMap. Then we update the IP field there and push it to the API.

I think this approach is better since it avoids a call to the Deployments endpoint. Also, it keeps the `newConfigMap` simpler, since it's only used for one usecase, returning a fresh CM.

I haven't benchmarked this, but I don't think this affects performance in any noticeable way.

@LiliC WDYT?

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -213,7 +213,7 @@ func (hc *HabitatController) onAdd(obj interface{}) {
 	level.Info(hc.logger).Log("msg", "created deployment", "name", d.GetObjectMeta().GetName())
 
 	// Create the ConfigMap for the peer watch file.
-	configMap := newConfigMap(sg.Name, d.UID, "")
+	configMap := newConfigMap(sg.Name, d.UID)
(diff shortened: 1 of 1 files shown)


LiliC:
I do not see the benefit of copying as when [we do a request to the Kubernetes API](https://github.com/kinvolk/habitat-operator/blob/19803351972f0494b088ff8b4fc41d39e45d2360/pkg/habitat/controller/controller.go#L355), [it creates a new struct](https://github.com/kinvolk/habitat-operator/blob/19803351972f0494b088ff8b4fc41d39e45d2360/vendor/k8s.io/client-go/kubernetes/typed/core/v1/configmap.go#L110), so we are not actually modifying the cache, hence we don't need to create a new copy of it. Unless I am missing something?

I agree on the saving the additional API call. Maybe if we were to read it from cache? But if we were to introduce the cache, we would have to also introduce the `workqueue` as we would have concurrent events.

asymmetric:
You're right, I was confusing the [advice](https://github.com/kubernetes/apiextensions-apiserver/blob/master/examples/client-go/controller/controller.go#L86-L93) not to modify objects from the store with our situation, in which we're getting objects straight from the API, bypassing any cache.

PR #48, opened 2017-08-14
Simplify Leader IP handling

asymmetric:
This PR simplifies the way we deal with the Leader IP file.

We can find out if the current leader is running using field selectors. This way, we only need to check if the length of the returned list of Pods is > 0, to know whether we need to do anything.

The PR also removes additional checks for the Pod status that are unnecessary since we already use fields to filter.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -285,12 +285,15 @@ func (hc *HabitatController) onPodUpdate(oldObj, newObj interface{}) {
 	// TODO: Do not retrieve or write IP if we are deploying a standalone topology.
 	pod, ok := newObj.(*apiv1.Pod)
 	if !ok {
-		level.Error(hc.logger).Log("msg", "Failed to cast pod.")
+		level.Error(hc.logger).Log("msg", "Unknown event received in Pod Update handler")
 		return
 	}
+
 	if pod == nil {
+		level.Warn(hc.logger).Log("msg", "Empty event received in Pod Update handler")
 		return
 	}
+
 	if pod.Status.Phase != apiv1.PodRunning {
 		return
 	}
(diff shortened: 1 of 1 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -356,21 +369,27 @@ func (hc *HabitatController) writeIP(pod *apiv1.Pod) error {
...
+	if currentLeaderIP != "" {
+		if ip == currentLeaderIP {
 			return nil
 		}
-		podList, err := hc.config.KubernetesClientset.CoreV1().Pods(apiv1.NamespaceDefault).List(metav1.ListOptions{})
+
+		// Is the leader still running?
+		// If so, we don't need to do anything.
+		fs := fields.SelectorFromSet(fields.Set{
+			"status.podIP": currentLeaderIP,
```
LiliC: In our `README.md` we claim that this project is compatible with Kubernetes `1.7.0` whereas this does not work as I described in the [issue](https://github.com/kubernetes/kubernetes/issues/50075) I created. The issue was addressed but it is NOT in any of the releases yet.

asymmetric: :sob:

asymmetric: I'd say we can then wait for 1.7.4 to be released, which will hopefully fix this.

LiliC: Sure we can leave the PR open, and when `1.7.4` is released we can revisit, wdyt? Only downside is that then we can't support `1.7.0`.

asymmetric: Don't particularly mind if we don't support < 1.7.4.

img

asymmetric:
Closed by #63.

PR #56, opened 2017-08-15
Check dependencies are present in Pod handler

### LINKED ISSUE #51: ConfigMap not found error
After deleting the SG with `kubectl`, the operator still receives events on the Pod handler.

In those handlers, we expect the ConfigMaps to be there, but they aren't (as they are deleted in the `onDelete`), so we get errors like 

```
level=error component=controller msg="configmaps \"example-encrypted-service-group\" not found"
```

asymmetric:
If either the Deployment or the ConfigMap are not present, the Pod
handler should exit early.

This happens because of the different rate at which resources are
deleted. If a SG is deleted, the Deployment and the ConfigMap are as
well.

But the Pod might take a little longer to be deleted, and in the
meantime, events for it will still be fired.

The handler should therefore check whether the resources it needs are
present, and if they aren't, it should simply exit.

Closes #51.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -318,7 +318,7 @@ func (hc *HabitatController) onPodDelete(obj interface{}) {
 	cmName := configMapName(sgName)
 	cm, err := hc.config.KubernetesClientset.CoreV1().ConfigMaps(apiv1.NamespaceDefault).Get(cmName, metav1.GetOptions{})
 	if err != nil {
-		level.Error(hc.logger).Log("msg", err)
+		level.Debug(hc.logger).Log("msg", "Pod event received, but ConfigMap already deleted")
 		return
 	}
 	currIP := cm.Data[peerFile]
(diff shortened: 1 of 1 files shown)


asymmetric:
I'm not super happy with this solution. It feels a bit hacky.

I've taken another look at [this](https://github.com/kubernetes/community/blob/master/contributors/devel/controllers.md) and the [ReplicaSet controller](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/replicaset/replica_set.go), but I'm not sure their solution applies to us.

LiliC:
The problem I see with this solution is that we will need to do this for all the cases in the future. Also we are again calling K8s API and all of this concurrently.

Yes, the general strategy would apply to us, I think we should introduce the work-queue. But we should also watch events from other resources, such as the `Pod`, `ConfigMap`,`Secret` and `Deployment`.  As right now we are acting only on `SG` and `Pod` events, whereas we should aim to be in desired state on each event, that can be anything from a pod deletion, secret being update, etc. This is where the work-queue comes in handy, we would process the items one by one instead of concurrently.

> The would solve the problem this PR tries to solve because if the SG has been deleted, the job handler wouldn't do anything.

Yes and no, we should still act on that event and be in the desired state. If we were to be using cache, all of these checks would be "free".

> Question: do we need to handle the seemingly complicated expectations dance?

I don't think so? I looked at other operators and they do not seem to do that.

asymmetric:
>  Yes and no, we should still act on that event and be in the desired state.

I'd argue that if the SG handler gets a job because a pod has been updated, but the corresponding SG object has already been removed, we don't really need to do anything. It's just a case of the two events arriving in the "wrong" order. Soon the pod will be deleted as well, so this update is not really relevant.

What would you do to reconciliate?

LiliC:
Yes, the cache would always be in the right state, so when reconciling we can check if this object still exists in cache, and delete the associated objects, no matter the order or type (for example `ServiceGroup` being deleted) of events.

LiliC:
@asymmetric Can this PR be closed?

asymmetric:
Superseded by #63.

Issue #54, opened 2017-08-15
Create e2e tests


LiliC:
Create e2e tests.

LiliC:
My initial conclusions on the tests we can do right now. So we do not every time when we make changes to our codebase, need to spin up our operator manually and create each example I would suggest starting with E2E tests.

E2e tests would be located in the `tests` directory and would cover scenarios such as resource deletion, checking whether we still have IP of a pod that is running, etc. At first we can run them locally against our `minikube`, maybe later on have a setup for those to be run automatically.

When we decide on our projects architecture we can start on focusing more on unit tests.

@asymmetric wdyt?

asymmetric:
I'm not sure I understand what the conclusions are :)

Is it that you'd like to focus on E2E tests  instead of unit testing the `writeIP` method? That's totally fine by me.

I would still keep this issue around for unit testing that method, and add a description in this issue of the 4 things we discussed yesterday.

LiliC:
Added the suggestions, feel free to edit, if you can think of something else, or if I misunderstood.

asymmetric:
Is this issue about E2E testing or unit testing? I think that could be reflected in the title/description.

LiliC:
Would you prefer if we keep this issue just for unit tests? And we create a separate one for e2e? I am fine with that.

asymmetric:
I'm just not sure what this issue is about, but I'm fine with you choosing that :)

PR #57, opened 2017-08-16
*: add building operator image

LiliC:
Operator already can run in-cluster, this automates building the image.

This PR is a first step towards adding e2e tests (https://github.com/kinvolk/habitat-operator/issues/54), as we need an image to run the test on a cluster.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,15 @@
+HUB :=
+REPO := kinvolk
+IMAGE := $(if $(HUB),$(HUB)/)$(REPO)/habitat-operator
+TAG := $(shell git describe --tags --always)
+
 build:
 	go build -i github.com/kinvolk/habitat-operator/cmd/operator
 
-.PHONY: build
+linux:
+	env GOOS=linux go build github.com/kinvolk/habitat-operator/cmd/operator
+
+image: linux
+	docker build -t "$(IMAGE):$(TAG)" .
+
+.PHONY: build linux image
--- a/README.md
+++ b/README.md
@@ -13,15 +13,31 @@ The Habitat Operator makes use of [`Custom Resource Definition`][crd]s, and requ
 
 ## Usage
 
-To run the `habitat-operator` as a binary outside of a Kubernetes cluster, run:
+### Running outside of the Kubernetes cluster:
+
+First build the `habitat-operator` binary by running:
+
+    make build
+
+This will produce a binary file, then start your operator by running:
 
     operator --kubeconfig ~/.kube/config
 
 To try out the operator with an example service, run:
 
-    kubectl create -f examples/habitat_service-standalone.yml
+    kubectl create -f examples/standalone
 
 This will create a 1-pod deployment of an `nginx` Habitat service.
+
+### Running inside of the Kubernetes cluster:
+
+First build the image:
+
+    make image
+
+This will produce a `kinvolk/habitat-operator` image, which can then be deployed to your cluster.
+
+The name of the generated docker image can be changed with an `IMAGE` variable, for example `make image IMAGE=mycorp/my-habitat-operator`. If the `habitat-operator` name is fine, then a `REPO` variable can be used like `make image REPO=mycorp` to generate the `mycorp/habitat-operator` image. Use the `TAG` variable to change the tag to something else (the default value is taken from `git describe --tags --always`) and a `HUB` variable to avoid using the default docker hub.
 
 ## Contributing
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,5 @@
+FROM alpine:3.6
+
+COPY operator /operator
+
+ENTRYPOINT ["/operator"]


Makefile
```diff
@@ -1,4 +1,14 @@
...
+
 build:
 	go build -i github.com/kinvolk/habitat-operator/cmd/operator
 
-.PHONY: build
+linux:
+	env GOOS=linux go build github.com/kinvolk/habitat-operator/cmd/operator
+
+image: linux
+	docker build -t "$(IMAGE):$(TAG)" .
```
LiliC: A developer might would want to build an image and push it to their own docker hub for example, so naming it something like `lili/habitat-operator`. I am fine with removing that, just thought would make things easier.

What do you mean by them being used in other targets?

Makefile
```diff
@@ -1,4 +1,14 @@
+REPO ?= kinvolk
+IMAGE ?= $(REPO)/habitat-operator
+TAG ?= $(shell git describe --tags --always)
```
krnowak: How do you invoke `make` with custom `REPO`, `IMAGE` and `TAG` variables? If you do `make REPO=mycorp` then using `:=` instead of `?=` should be enough. And I would prefer it that way, because now these variables can be accidentally overridden if you have REPO env var. Also, the process of building the image should be documented - do we have some README, where we could put this?

Also, maybe consider adding the `HUB` variable which is empty by default. So I can override it with `docker.io` on my fedora to avoid using fedora-specific docker hub. You could use it in the `IMAGE` variable like this: `IMAGE := $(if $(HUB),$(HUB)/)$(REPO)/habitat-operator`.

Makefile
krnowak: @asymmetric probably means using the `IMAGE` variable in `github.com/$(IMAGE)/cmd/operator`. I'd prefer to avoid conflating the `IMAGE` variable as two things - as a docker image name and as a github repo name.

I think that the `go build` invocations should not use `github.com/kinvolk/habitat-operator` at all, but rather be `go build ./cmd/operator`. But that would need testing if it works and should be a part of another PR.

Makefile
LiliC: Done. Can you have another look? Thanks.

README.md
```diff
@@ -13,16 +13,30 @@ The Habitat Operator makes use of [`Custom Resource Definition`][crd]s, and requ
...
 To try out the operator with an example service, run:
 
-    kubectl create -f examples/habitat_service-standalone.yml
+    kubectl create -f examples/standalone
 
 This will create a 1-pod deployment of an `nginx` Habitat service.
 
+### Running inside of the Kubernetes cluster:
+
+To run the `habitat-operator` inside of a Kubernetes cluster, first build the image:
```
asymmetric: Don't think you need to repeat the title here :)

README.md
```diff
@@ -13,16 +13,30 @@ The Habitat Operator makes use of [`Custom Resource Definition`][crd]s, and requ
...
+
+    make image
+
+This will produce a `kinvolk/habitat-operator` image, which can then be deployed to your cluster.
```
asymmetric: Should we mention how people can use `REPO` and the other vars?

README.md
LiliC: Done, can you have another look, thanks.

README.md
```diff
@@ -13,16 +13,30 @@ The Habitat Operator makes use of [`Custom Resource Definition`][crd]s, and requ
...
 
 This will create a 1-pod deployment of an `nginx` Habitat service.
 
+### Running inside of the Kubernetes cluster:
+
+First build the image:
+
+    make image
+
+This will produce a `kinvolk/habitat-operator` image, which can then be deployed to your cluster.
```
krnowak: How about adding:

The name of the generated docker image can be changed with an `IMAGE` variable like so `make image IMAGE=mycorp/my-habitat-operator`. If the `habitat-operator` name is fine, then a `REPO` variable can be used like `make image REPO=mycorp` to generate the `mycorp/habitat-operator` image. Use a `TAG` variable to change the tag to something else (default value is taken from `git describe --tags --always`) and a `HUB` variable to avoid using the default docker hub.

README.md
asymmetric: :(

LiliC: No need for :(, will add this https://github.com/kinvolk/habitat-operator/pull/57#discussion_r133660119, wdyt?

README.md
```diff
@@ -33,6 +32,8 @@ This will create a 1-pod deployment of an `nginx` Habitat service.
 
 First build the image:
 
+The name of the generated docker image can be changed with an `IMAGE` variable, for example `make image IMAGE=mycorp/my-habitat-operator`. If the `habitat-operator` name is fine, then a `REPO` variable can be used like `make image REPO=mycorp` to generate the `mycorp/habitat-operator` image. Use a `TAG` variable to change the tag to something else (default value is taken from `git describe --tags --always`) and a `HUB` variable to avoid using the default docker hub.
+
```
asymmetric: the default. And I would say "use *the* TAG variable" instead of *a*.

README.md
asymmetric: :)

LiliC:
Address your comments @asymmetric @krnowak can you have another look.

README.md
```diff
@@ -13,16 +13,30 @@ The Habitat Operator makes use of [`Custom Resource Definition`][crd]s, and requ
...
 To try out the operator with an example service, run:
 
-    kubectl create -f examples/habitat_service-standalone.yml
+    kubectl create -f examples/standalone
 
 This will create a 1-pod deployment of an `nginx` Habitat service.
 
+### Running inside of the Kubernetes cluster:
+
+First build the image. The name of the generated docker image can be changed with an `IMAGE` variable, for example `make image IMAGE=mycorp/my-habitat-operator`. If the `habitat-operator` name is fine, then a `REPO` variable can be used like `make image REPO=mycorp` to generate the `mycorp/habitat-operator` image. Use the `TAG` variable to change the tag to something else (the default value is taken from `git describe --tags --always`) and a `HUB` variable to avoid using the default docker hub.
```
krnowak: I'd put this paragraph below the "This will prodice a `kinvolk/habitat-operator` image, …".

krnowak:
A single nit, otherwise fine to merge.

PR #58, opened 2017-08-17
Log error when Deployment can't be created due to missing Secret

asymmetric:
It's an error so we should use `Error`.

Also, I think we don't need to check the resource for `== nil`, as I don't think Kubernetes would deliver us a `nil` struct if there wasn't an `error`.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -231,9 +231,6 @@ func (hc *HabitatController) onPodUpdate(oldObj, newObj interface{}) {
 		level.Error(hc.logger).Log("msg", "Failed to cast pod.")
 		return
 	}
-	if pod == nil {
-		return
-	}
 	if pod.Status.Phase != apiv1.PodRunning {
 		return
 	}
@@ -248,9 +245,6 @@ func (hc *HabitatController) onPodDelete(obj interface{}) {
 	pod, ok := obj.(*apiv1.Pod)
 	if !ok {
 		level.Error(hc.logger).Log("msg", "Failed to cast pod.")
-		return
-	}
-	if pod == nil {
 		return
 	}
 	sgName, exists := pod.ObjectMeta.Labels["service-group"]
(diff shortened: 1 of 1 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -402,20 +402,15 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
 		// Let's make sure our secret is there before mounting it.
 		secret, err := hc.config.KubernetesClientset.CoreV1().Secrets(apiv1.NamespaceDefault).Get(sg.Spec.Habitat.Config, metav1.GetOptions{})
 		if err != nil {
-			return nil, err
-		}
-
-		if secret == nil {
-			// If we can't find the secret we should just return the base deployment.
-			level.Warn(hc.logger).Log("msg", "Secret could not be mounted as it did not exist.")
+			level.Error(hc.logger).Log("msg", "Could not find Secret containing service configuration")
```
LiliC: Can we maybe mention the Secret name in the log?

asymmetric:
PTALA.

pkg/habitat/controller/controller.go
```diff
@@ -402,20 +396,15 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
 		// Let's make sure our secret is there before mounting it.
 		secret, err := hc.config.KubernetesClientset.CoreV1().Secrets(apiv1.NamespaceDefault).Get(sg.Spec.Habitat.Config, metav1.GetOptions{})
 		if err != nil {
-			return nil, err
-		}
-
-		if secret == nil {
-			// If we can't find the secret we should just return the base deployment.
-			level.Warn(hc.logger).Log("msg", "Secret could not be mounted as it did not exist.")
+			level.Error(hc.logger).Log("msg", "Could not find Secret containing service configuration", "name", sg.Spec.Habitat.Config)
```
LiliC: Looking at it again, I don't think we need to log this. As we already return an error and later on log it:
https://github.com/kinvolk/habitat-operator/blob/1a3261ebbeffcc1dbcc030371571213d544ff5ce/pkg/habitat/controller/controller.go#L146

wdyt?

asymmetric:
PTALA.

PR #59, opened 2017-08-17
Add log/level to Gopkg.lock

asymmetric:
This was added after running `dep ensure` again.

This should've been there before already, not really sure how things were working :)

Maybe it's a new version of `dep` adding this?

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -39,7 +39,7 @@
 
 [[projects]]
   name = "github.com/go-kit/kit"
-  packages = ["log"]
+  packages = ["log","log/level"]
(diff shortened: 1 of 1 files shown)


LiliC:
Not sure what happened there. But I noticed in https://github.com/kinvolk/habitat-operator/pull/63 you have https://github.com/kinvolk/habitat-operator/pull/63/commits/7d08051c5a72686ec5a6967d4d97274a3e34d045 Do you want to close this one or remove the commit there?

asymmetric:
Let's merge this and see how hard it is to resolve the conflict :)

PR #61, opened 2017-08-18
Add make target for unit tests

asymmetric:
I think `test` as a target makes sense now, and when we will merge E2E tests can rename this to `unit` and use `test` for both.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,4 +12,7 @@ linux:
 image: linux
 	docker build -t "$(IMAGE):$(TAG)" .
 
-.PHONY: build linux image
+test:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -12,4 +12,7 @@ linux:
 image: linux
 	docker build -t "$(IMAGE):$(TAG)" .
 
-.PHONY: build linux image
+test:
```
krnowak: It just occurred to me that we don't use the well-known nouns in our Makefile (all, check, clean…). Ah, well - probably not really important if we have this stuff documented.

PR #63, opened 2017-08-18
Implement workqueue

### LINKED ISSUE #26: Introduce workqueue
To make sure individual events don't interfere with each other, upstream Kubernetes suggestion is for controllers to implement a workqueue. 

More info can be found [here.](https://github.com/kubernetes/community/blob/master/contributors/devel/controllers.md)

asymmetric:
This PR adds a ServiceGroup workqueue.

Following the guidelines, all event handlers (SG and Pod for now) do is retrieve the appropriate SG key and add a job to the queue.

The worker then consumes these jobs, and performs the necessary operations.

Closes #26, #51 and obsoletes #56.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -24,6 +24,7 @@ import (
 	"github.com/go-kit/kit/log/level"
 	appsv1beta1 "k8s.io/api/apps/v1beta1"
 	apiv1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
@@ -32,6 +33,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
 
 	crv1 "github.com/kinvolk/habitat-operator/pkg/habitat/apis/cr/v1"
 )
@@ -45,6 +47,14 @@ const (
 type HabitatController struct {
 	config Config
 	logger log.Logger
+
+	// queue contains the jobs that will be handled by syncServiceGroup.
+	// A workqueue.RateLimitingInterface is a queue where failing jobs are re-enqueued with an exponential
+	// delay, so that jobs in a crashing loop don't fill the queue.
+	queue workqueue.RateLimitingInterface
+
+	// store is the cache of ServiceGroups retrieved by the ListWatcher.
+	store cache.Store
 }
 
 type Config struct {
@@ -70,6 +80,7 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
 	hc := &HabitatController{
 		config: config,
 		logger: logger,
+		queue:  workqueue.NewNamedRateLimitingQueue(workqueue.DefaultControllerRateLimiter(), "servicegroup"),
 	}
 
 	return hc, nil
@@ -79,9 +90,12 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
 func (hc *HabitatController) Run(ctx context.Context) error {
 	level.Info(hc.logger).Log("msg", "Watching Service Group objects")
 
-	hc.watchCustomResources(ctx)
+	hc.watchServiceGroups(ctx)
 
 	hc.watchPods(ctx)
+
+	// Start the synchronous queue consumer.
+	go hc.worker()
 
 	// This channel is closed when the context is canceled or times out.
 	<-ctx.Done()
@@ -90,14 +104,14 @@ func (hc *HabitatController) Run(ctx context.Context) error {
 	return ctx.Err()
 }
 
-func (hc *HabitatController) watchCustomResources(ctx context.Context) {
+func (hc *HabitatController) watchServiceGroups(ctx context.Context) {
 	source := cache.NewListWatchFromClient(
 		hc.config.HabitatClient,
 		crv1.ServiceGroupResourcePlural,
 		apiv1.NamespaceAll,
 		fields.Everything())
 
-	_, k8sController := cache.NewInformer(
+	store, k8sController := cache.NewInformer(
 		source,
 
 		// The object type.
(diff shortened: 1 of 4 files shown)


asymmetric:
@LiliC are your E2E test going to cover the functionality of the peer IP handling, e.g. that we correctly replace the IP of a dead pod?

LiliC:
@asymmetric My idea was we just need to test that a CM was updated after our Pod died, not sure we need to care about if it was the correct IP. Think that might be more unit tests, wdyt?

pkg/habitat/controller/controller.go
```diff
@@ -110,79 +120,168 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) {
...
+
+func (hc *HabitatController) getRunningPods(namespace string) ([]apiv1.Pod, error) {
+	fs := fields.SelectorFromSet(fields.Set{
+		"status.phase": "Running",
+	})
+	running := metav1.ListOptions{
+		FieldSelector: fs.String(),
+	}
+
+	// TODO We can use the store for this query.
```
LiliC: I would suggest we do this now. From what I can see we already have access to pod store/cache, we just need to save it? Or is there anything else stopping us?

pkg/habitat/controller/controller.go
```diff
@@ -225,104 +325,59 @@ func (hc *HabitatController) onPodAdd(obj interface{}) {
...
-	pod, ok := newObj.(*apiv1.Pod)
-	if !ok {
-		level.Error(hc.logger).Log("msg", "Failed to cast pod.")
+	old, ok1 := oldObj.(*apiv1.Pod)
+	if !ok1 {
+		level.Error(hc.logger).Log("msg", "Failed to type assert pod", "obj", oldObj)
+		return
+	}
+
+	new, ok2 := newObj.(*apiv1.Pod)
```
LiliC: `new` maybe not a reserved name, but I would still like to avoid using that. WDYT?

pkg/habitat/controller/controller.go
```diff
@@ -428,6 +483,113 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
...
+	// * an SG was updated
+	// * a Pod was created
+	// * a Pod was deleted
+	// * a Pod was updated
+	obj, exists, err := hc.store.GetByKey(key)
+	if err != nil {
+		return err
+	}
+	if !exists {
+		// The SG was deleted.
```
LiliC: Ignore the comment, the `store` naming confused me.

pkg/habitat/controller/controller.go
asymmetric: Didn't know about [this](https://golang.org/ref/spec#Allocation).

pkg/habitat/controller/controller.go
asymmetric: I was thinking this and other things (like increasing the number of workers) can be done as additional PRs later on, as there is already enough going on in this PR.

pkg/habitat/controller/controller.go
```diff
@@ -428,6 +483,113 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
...
+
+		return true
+	}
+
+	hc.queue.Forget(key)
+
+	return true
+}
+
+func (hc *HabitatController) syncServiceGroup(key string) error {
```
asymmetric: This follows the example of [other](https://github.com/kubernetes/kubernetes/blob/e633a1604f00908a1dcc898b206c3404db4d82ed/pkg/controller/service/service_controller.go#L716) [controllers](https://github.com/kubernetes/kubernetes/blob/e633a1604f00908a1dcc898b206c3404db4d82ed/pkg/controller/replicaset/replica_set.go#L543), and I guess the idea is that this function syncs the resource this controller cares about, including all depdencies.

pkg/habitat/controller/controller.go
```diff
@@ -428,6 +483,113 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
...
+	// we got woken up, so either:
```
asymmetric: That slipped in, wasnt' planning on having it committed :)

But I guess if it's reworded it can be useful, as you say.

pkg/habitat/controller/controller.go
```diff
@@ -45,6 +47,10 @@ const (
 type HabitatController struct {
 	config Config
 	logger log.Logger
+
+	queue workqueue.RateLimitingInterface
+
+	store cache.Store
```
asymmetric: For now it's the only store though :)

pkg/habitat/controller/controller.go
```diff
@@ -110,79 +120,168 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) {
 
 		// Your custom resource event handlers.
 		cache.ResourceEventHandlerFuncs{
-			AddFunc:    hc.onAdd,
-			UpdateFunc: hc.onUpdate,
-			DeleteFunc: hc.onDelete,
+			AddFunc: hc.enqueueSG,
+			UpdateFunc: func(old, new interface{}) {
+				oldSG, ok1 := old.(*crv1.ServiceGroup)
```
asymmetric: We can do that even without extracting to its own function though. I quite like this small anonymous function (and I stole it from [here](https://github.com/kubernetes/kubernetes/blob/e633a1604f00908a1dcc898b206c3404db4d82ed/pkg/controller/service/service_controller.go#L131) ;))

pkg/habitat/controller/controller.go
LiliC: Maybe, but it still stores just SG cache so we can still rename it to describe what it stores. Its up to you I guess.

pkg/habitat/controller/controller.go
LiliC: I am fine with that, but then we shouldn't close the issue it relates to, but rather maybe add a list of AC to the issue. Otherwise we can forget about it.  

Something like:
- Introduce WorkQueue
- Add Deployment watching
- Store Pod cache

etc.
This was we can just pick up those tasks and keep track of them.

pkg/habitat/controller/controller.go
asymmetric: I think I'll leave it like this, if you don't mind.

pkg/habitat/controller/controller.go
asymmetric: If you mean #51, this PR does close that issue AFAICT. The features you mentioned are additional nice-to-haves IMO.

Or did you mean another issue?

LiliC:
I tested this branch and it seems like writing IP is not working the same, compared to master.

When I create two SG one after another, the ConfigMaps contain the same IP.

pkg/habitat/controller/controller.go
LiliC: Issue https://github.com/kinvolk/habitat-operator/issues/26.

asymmetric: I'll add separate issues for the items you mentioned above.

pkg/habitat/controller/controller.go
```diff
@@ -110,79 +120,177 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) {
...
 		cache.ResourceEventHandlerFuncs{
-			AddFunc:    hc.onAdd,
-			UpdateFunc: hc.onUpdate,
-			DeleteFunc: hc.onDelete,
+			AddFunc: hc.enqueueSG,
+			UpdateFunc: func(oldObj, newObj interface{}) {
+				oldSG, ok1 := oldObj.(*crv1.ServiceGroup)
+				newSG, ok2 := newObj.(*crv1.ServiceGroup)
+
+				if !ok1 {
```
LiliC: Maybe move it before `newSG, ok2`? That way we can return early and not create `newSG`.

asymmetric:
Good catch! PTALA.

pkg/habitat/controller/controller.go
```diff
@@ -110,79 +124,181 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) {
...
+
+		return nil
+	}
+
+	// There are running Pods, add the IP of one of them to the ConfigMap.
+	leaderIP := runningPods[0].Status.PodIP
+
+	newCM := newConfigMap(sg.Name, deploymentUID, leaderIP)
+
+	var cm *apiv1.ConfigMap
```
LiliC: Why is this declaration needed?

pkg/habitat/controller/controller.go
```diff
@@ -110,79 +124,181 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) {
...
+func (hc *HabitatController) getRunningPods(namespace, label string) ([]apiv1.Pod, error) {
+	fs := fields.SelectorFromSet(fields.Set{
+		"status.phase": "Running",
+	})
+	ls := fields.SelectorFromSet(fields.Set(map[string]string{
+		crv1.ServiceGroupLabel: label,
+	}))
+
+	running := metav1.ListOptions{
+		FieldSelector: fs.String(),
```
LiliC: I would just create `fs` and `ls` inline, as it more readable.

pkg/habitat/controller/controller.go
```diff
@@ -110,79 +124,181 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) {
...
+		return err
 	}
 
-	d, err := hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(apiv1.NamespaceDefault).Create(deployment)
-	if err != nil {
-		level.Error(hc.logger).Log("msg", err)
-		return
+	// Create Deployment, if it doesn't already exist.
+	d, dErr := hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(apiv1.NamespaceDefault).Create(deployment)
+	if dErr != nil {
```
LiliC: Why not the standard naming of `error`?

pkg/habitat/controller/controller.go
```diff
@@ -110,79 +124,181 @@ func (hc *HabitatController) watchCustomResources(ctx context.Context) {
...
+		if !apierrors.IsAlreadyExists(dErr) {
+			return dErr
+		}
+
+		d, dErr = hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(sg.Namespace).Get(deployment.Name, metav1.GetOptions{})
```
LiliC: I am confused why we need to `Get`, as our `Create` already returns a deployment.

asymmetric: Because if we're here, the `Create` has failed.

LiliC: Hmm, I would suggest we restructure this, as right now from first glance its not really clear whats happening.

I would actually reverse those and start by first seeing if our deployment already exists, rather then trying to create it. If it doesn't then create it.

asymmetric: This way we save one call though. If the resource doesn't exist, we just create it, with no need perform the check first.

I've seen the pattern in [other](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/service/service_controller.go#L325) [controllers](https://github.com/coreos/prometheus-operator/blob/3b3eb277a253e725158d76e008c8a92164ba6fd7/pkg/prometheus/operator.go#L671) as well, so I'm more inclined on leaving this as is.

asymmetric: The performance argument will probably become irrelevant if/when we move to using a `Deployment`s cache, so I think we can refactor this there.

pkg/habitat/controller/controller.go
asymmetric: I find this style more readable, and unless there are guidelines going against it, I'd like to keep it.

It's also how we do it [elsewhere](https://github.com/kinvolk/habitat-operator/blob/84b08acfb61e10628ca0b110e1c2677f2f527e0c/pkg/habitat/controller/controller.go#L321).

asymmetric:
Are there any more blockers? If not, can this get a green-light?

asymmetric:
I've created two issues: #65, #66.

PR #70, opened 2017-08-23
Use namespace from resource instead of default

### LINKED ISSUE #50: Add support for custom namespace
We should allow the user to create the Custom Object in a namespace of their choosing.

If none is specified, it will be created in the default namespace.

asymmetric:
This allows us to support custom namespaces for ServiceGroups and all
resources handled by the operator.

I'm not even sure we need an example, as that's more part of how K8s works. WDYT?

Closes #50.

### DIFF
--- a/examples/namespaced/README.md
+++ b/examples/namespaced/README.md
@@ -1,0 +1,8 @@
+# Namespaced ServiceGroup example
+
+## Workflow
+
+Simply run `kubectl create -f examples/namespaced/service_group.yml`.
+
+Note that any Secrets used by a ServiceGroup must be in the same namespace as
+the ServiceGroup itself.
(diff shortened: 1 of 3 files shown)


LiliC:
Just wondering what if there is no `Namespace` created, does it default to `default namespace`?

asymmetric:
Yes, this is handled transparently by Kubernetes.

LiliC:
Cool!

One more thing. Was thinking about our secret, how should we handle that? Maybe we should instead just use `NamespaceAll` there? Or do we expect the user to also deploy the secret to same `Namespace` as the `SG`? WDYT?

https://github.com/kinvolk/habitat-operator/blob/1084ff781325650fcbf12954dbfef503d0cbe514/pkg/habitat/controller/controller.go#L475

asymmetric:
I think the secret should be in the same NS as the SG. I've updated the code and added a README mentioning this.

PR #73, opened 2017-08-23
Handle leader-follower topology

### LINKED ISSUE #24: Add support for --peer-watch-file
The supervisor should be started with the `--peer-watch-file` flag, if the CRD had a key `topology: leader`.

The flag should be passed as an argument to the container.

### LINKED ISSUE #25: Add support for --topology flag
The operator needs to start the supervisor with the `--topology leader` flag, if the spec has a `topology: leader` flag.

This depends on https://github.com/habitat-sh/habitat/issues/2735, because we want to start all supervisors with the same set of flags.

asymmetric:
With this PR, the operator is able to handle the creation of Service Groups in a [leader-follower topology](https://www.habitat.sh/docs/run-packages-topologies/).

We do this by using a yet unreleased version of the habitat supervisor that has a [`--peer-watch-file` flag](https://github.com/habitat-sh/habitat/issues/2735).
This file is populated by the operator with the IP of a running pod, which acts as the entrypoint to the Service Group.

The habitat supervisor watches this file for changes, and when it detects them, restarts the application.

Closes #24.
Closes #25.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,15 @@
 
 The Habitat Operator makes use of [`Custom Resource Definition`][crd]s, and requires a Kubernetes cluster of version `>= 1.7.0`.
 
+At the moment, the Operator requires a forked version of the Habitat supervisor
+that adds support for the `--peer-watch-file` flag.
+See [this issue](https://github.com/habitat-sh/habitat/issues/2735) to track
+progress on upstreaming the feature.
+
+This means that images need to be created using the aforementioned fork of the
+`hab` client, which can be accomplished using [this
+script](https://gist.github.com/krnowak/3c854e94245e2f33a8366e629bfb09c8).
+
 ## Installing
 
     go get -u github.com/kinvolk/habitat-operator/cmd/operator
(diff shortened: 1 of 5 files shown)


asymmetric:
@krnowak does the description of the hab side look correct?

krnowak:
> This file is populated by the operator with the IP of a running pod, which acts as the entrypoint to the Service Group.

as an entrypoint to the ring.

> The habitat supervisor watches this file for changes, and when it detects them, restarts the application.

and when it detects them, it tries to join the ring. When the service/application is started depends on the topology. If it's standalone, then the service can be started immediately (without waiting to join the ring, even). If it's leader-follower, then it waits for the election to finish, where the leader will be chosen and then starts the service. For the election to start we need quorum (which is I think 3 members) and to reach that quorum, we need supervisors to join the ring.

LiliC:
> We do this by using a yet unreleased version of the habitat supervisor that has a --peer-watch-file flag.

I believe this PR be a good place to add the warning in the `README`, about that. WDYT?

asymmetric:
PTALA at the added text in the README. I had a hard time figuring out how to explain this, so suggestions very welcome.

/cc @krnowak

PR #74, opened 2017-08-24
Add --verbose flag to control log level

### LINKED ISSUE #64: Allow changing log level from command line
Currently, a user needs to change the [source](https://github.com/kinvolk/habitat-operator/blob/1a7db57fabc352c00748dc31f3289618830cc0c0/cmd/operator/main.go#L42) to change the log level.

We should expose a flag that does this instead.

asymmetric:
This PR introduces a `--verbose` boolean flag that controls the log level.

When it's set, the level is `DEBUG`, otherwise it's `INFO`.

At the moment, I don't think we need a more granular control of the log level.
 
Closes #64.

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -38,12 +38,19 @@ type Config struct {
 }
 
 func run() int {
-	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
-	logger = level.NewFilter(logger, level.AllowInfo())
-
 	// Parse config flags.
 	kubeconfig := flag.String("kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	verbose := flag.BoolP("verbose", "v", false, "Enable verbose logging.")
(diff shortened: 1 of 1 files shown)


cmd/operator/main.go
```diff
@@ -38,13 +38,20 @@ type Config struct {
 }
 
 func run() int {
-	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
-	logger = level.NewFilter(logger, level.AllowInfo())
-
 	// Parse config flags.
 	kubeconfig := flag.String("kubeconfig", "", "Path to a kubeconfig. Only required if out-of-cluster.")
+	verbose := flag.BoolP("verbose", "v", false, "Enable verbose logging.")
```
LiliC: Just a suggestion, WDYT of `debug` instead of `verbose`.

asymmetric: @krnowak wdyt?

PR #77, opened 2017-08-24
Skip ConfigMap handling if topology is not leader-follower

### LINKED ISSUE #49: Do not handle leader IP in standalone topology
The Pod handlers should not perform any of the leader IP work if the Topology of the ServiceGroup is standalone.

asymmetric:
This affects the watcher, in that we are only interested in watching pods if there is work to do in relation to the ConfigMap, and it affects the `syncServiceGroup` method, because it looks for Pods as part of its ConfigMap handling logic.

Closes #49.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -437,6 +437,12 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
 	// is set, habitat by default sets standalone topology.
 	topology := crv1.TopologyStandalone
 
+	// These two variables relate to the mounting of the peer-ip ConfigMap, and
+	// are only used by the leader-follower topology, so we leave them blank for
+	// the default case.
+	var volumeMounts []apiv1.VolumeMount
+	var volumes []apiv1.Volume
+
 	if sg.Spec.Habitat.Topology == crv1.TopologyLeader {
 		topology = crv1.TopologyLeader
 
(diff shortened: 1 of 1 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -338,7 +340,13 @@ func (hc *HabitatController) handleServiceGroupDeletion(key string) error {
 }
 
 func (hc *HabitatController) watchPods(ctx context.Context) {
-	ls := labels.SelectorFromSet(labels.Set(map[string]string{"habitat": "true"}))
+	ls := labels.SelectorFromSet(labels.Set(map[string]string{
+		crv1.HabitatLabel: "true",
```
LiliC: Do we still need this habitat label, as we have the topology label now?

pkg/habitat/controller/controller.go
```diff
@@ -338,7 +340,13 @@ func (hc *HabitatController) handleServiceGroupDeletion(key string) error {
...
+		// We are only interested in Pods in a leader topology, since they are the
```
LiliC: I would simplify this comment and move it above right before creation of `ls`.

pkg/habitat/controller/controller.go
asymmetric: Yeah, not sure. I guess you could argue that this label is for finding out *any* habitat resource.

Maybe we could achieve the same by using a set selector on the topology label, like `topology in (leader, standalone`), but I guess using habitat is more explicit.

pkg/habitat/controller/controller.go
asymmetric: How would you simplify it? If we leave the other label as well, I think the comments makes sense here.

LiliC: Just my personal preference, think code is more readable if the comment is not placed in between, breaks the flow when reading code. But feel free to leave it here.

Maybe something like:
// Filter only pods with leader topology, as those are the ones that are started with the --peer-watch-file flag.

asymmetric: OK, then you need to approve this or change it to a comment :)

krnowak:
Should we wait with merging this until the meeting today?

asymmetric:
Closed, see #49.

PR #80, opened 2017-08-28
Fix topology in encrypted ring example

asymmetric:


asymmetric:
I think having leader-follower there makes more sense, from a grammatical POV, and also it makes it more clear what the topology *does*.

PR #81, opened 2017-08-28
Fix ring name passed as flag argument

asymmetric:
The argument we were passing included the revision, whereas habitat only
expects the ring name.

* Before: `--ring foobar-12312313`
* After: `--ring foobar`

### DIFF
--- a/examples/encrypted/service_group.yml
+++ b/examples/encrypted/service_group.yml
@@ -1,14 +1,15 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: ring-key-20170816102049
+  # the name must mirror the ring filename, without the extension
+  name: example-encrypted-ring-20170829113029
   labels:
     habitat: "true"
     ring-key: "true"
 type: Opaque
 data:
   # base64-encoded ring key
-  ring-key: U1lNLVNFQy0xCnJpbmcta2V5LTIwMTcwODE2MTAyMDQ5CgpyTmtKbkJxZHppVWdmdnYweEVhaWpzMWxHUGZUdHBLdFg0L0xaMmpQcWdJPQo=
+  ring-key: U1lNLVNFQy0xCmV4YW1wbGUtZW5jcnlwdGVkLXJpbmctMjAxNzA4MjkxMTMwMjkKClFGZm9ZTHJOV3NNRk93Tk4wb0F1d3d4WjA1aVBMdTQxSUdGSzJISVFqTlk9
 ---
 apiVersion: habitat.sh/v1
 kind: ServiceGroup
@@ -21,4 +22,4 @@ spec:
   habitat:
     topology: leader
     # the name of the secret containing the ring key
-    ringKey: ring-key-20170816102049
+    ringSecretName: example-encrypted-ring-20170829113029
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"regexp"
 	"time"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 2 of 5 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -536,18 +537,24 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
...
 			level.Error(hc.logger).Log("msg", "Could not find Secret containing ring key")
 			return nil, err
 		}
 
 		// The filename under which the ring key is saved.
-		ringKeyFile := fmt.Sprintf("%s.%s", ringName, ringKeyFileExt)
+		ringKeyFile := fmt.Sprintf("%s.%s", ringSecretName, ringKeyFileExt)
+
+		// The name of the ring, without the revision.
+		ringName := strings.Split(ringSecretName, "-")[0]
```
krnowak: In the example above, you use "ring-key-20170816102049", so this line will result in "ring", not "ring-key". I think you could do something like:
```
ringNameParts := strings.Split(str, "-")
if len(ringNameParts) < 2 {
	return nil, fmt.Errorf("malformed secret name: %s", ringSecretName)
}
ringName := strings.Join(ringNameParts[0:len(ringNameParts)-1], "-")
```

I don't know if the ring key revision should always be a date, but if so, other way of getting the ring name would be to use regex (I know, I know, now you would have two problems) like `^(.*)-\d{14}$`, so the name would be captured and the revision would be validated.

pkg/habitat/controller/controller.go
```diff
@@ -536,18 +537,24 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
...
+		if ringName == "" {
```
krnowak: That would happen only if ringSecretName is an empty string, so a key like "a" would not trigger this error.

pkg/habitat/controller/controller.go
```diff
@@ -536,18 +537,24 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
...
+			return nil, fmt.Errorf("malformatted secret name: %s", ringSecretName)
```
krnowak: malformed

examples/encrypted/service_group.yml
```diff
@@ -1,6 +1,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
+  # the name must contain the ring key revision as well
```
LiliC: `the name must contain the ring key revision as well` Besides what?
Maybe something like, 
> the name of the Secret must start with `ring-key-` and contain the ring key revision as well


Or maybe that belongs more in the docs, WDYT? https://github.com/kinvolk/habitat-operator/blob/258d1cee55f2d8d1500d0f00c477a0078ee8fe75/examples/encrypted/README.md

krnowak: Besides the name, so the name needs to be in format like `<name>-<revision>`.

asymmetric:
PTALA.

pkg/habitat/controller/utils.go
```diff
@@ -48,6 +48,16 @@ func validateCustomObject(sg crv1.ServiceGroup) error {
 		return fmt.Errorf("unkown topology: %s", spec.Habitat.Topology)
 	}
 
+	if rsn := spec.Habitat.RingSecretName; rsn != "" {
+		ringParts := ringRegexp.FindStringSubmatch(rsn)
+
+		// The ringParts slice should have a second element for the capturing group
+		// in the ringRegexp regular expression, containing the ring's name.
+		if len(ringParts) < 2 {
+			return fmt.Errorf("malformed secret name: %s", rsn)
```
LiliC: Since a user can potentially create multiple Secrets (one for initial config and one for ring-key), we should extend this error to include that this relates to the ring key secret. WDYT?

PR #82, opened 2017-08-29
Minor fixes and improvments

LiliC:
This PR:
- adds a timestamp to our logs, thought it would be valuable, when trying things out locally, I find it very useful to see when things were created, deleted, etc.

- removes `buildConfig ` as that check is already done in `BuildConfigFromFlags` function.
https://github.com/kinvolk/habitat-operator/blob/57b46f7cc225c87434432e6ccec1221f34eeb8e1/vendor/k8s.io/client-go/tools/clientcmd/client_config.go#L527-L534

### DIFF
--- a/cmd/operator/main.go
+++ b/cmd/operator/main.go
@@ -44,6 +44,7 @@ func run() int {
 
 	// Set up logging.
 	logger := log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr))
+	logger = log.With(logger, "ts", log.DefaultTimestamp)
 
 	if *verbose {
 		logger = level.NewFilter(logger, level.AllowDebug())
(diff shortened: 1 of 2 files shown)


cmd/operator/main.go
```diff
@@ -53,7 +53,7 @@ func run() int {
 	}
 
 	// Build operator config.
-	config, err := buildConfig(*kubeconfig)
+	config, err := clientcmd.BuildConfigFromFlags("", *kubeconfig)
```
asymmetric: Good catch. You might want to open a PR to fix [this example in client-go](https://github.com/kubernetes/apiextensions-apiserver/blob/79ecda8df91cd9304503d6f3e488341eabe2287f/examples/client-go/main.go#L125).

PR #86, opened 2017-08-30
Only run minikube in e2e target

asymmetric:
This removes the error we had when running `make` and `minikube` was not running/present.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-EIP := $(shell minikube ip)
 HUB :=
 REPO := kinvolk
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -18,6 +17,7 @@ test:
 	go test -v $(shell go list ./... | grep -v /vendor/ | grep -v /test/)
 
 e2e:
+	@: ${EIP := $(shell minikube ip)}
```
krnowak: This won't work. Instead of adding this line here, just do `"$$(minikube ip)"` instead of `"$(EIP)"` two lines below.

asymmetric:
PTALA.

Issue #76, opened 2017-08-24
Run E2E tests automatically


asymmetric:
This issue is to explore ways in which we can make running our E2E tests part of the CI process.

* Parallelizing Travis jobs using the [build matrix](https://docs.travis-ci.com/user/speeding-up-the-build/)
* Using the `-j` flag in `make`
* Using a [cron job](https://docs.travis-ci.com/user/cron-jobs/) to periodically run the E2E tests on master
* Using a custom bash script that only runs the E2E tests on master
* Only running the E2E tests in the job that tests the PR merge commit (e.g. checking `$TRAVIS_PULL_REQUEST` in a bash script)

LiliC:
Two options I see for for running our e2e tests on Travis CI: 
- We could use `minikube` for this, there is a flag to use no drivers `--vm-driver=none`. That should bypass the problem with having to get virtual machines running on top of Travis CI. Using `minikube` gives us the option to reproduce the failed tests locally and therefor fix them much easily.
- There is also [kubeadm-dind-cluster](https://github.com/Mirantis/kubeadm-dind-cluster)

PR #90, opened 2017-09-06
*: setup e2e tests on Travis CI

LiliC:
This PR enables running end-to-end tests on Travis CI. This sets up Kubernetes with minikube on Travis CI, with the help of flag `--vm-driver=none`.

Closes https://github.com/kinvolk/habitat-operator/issues/76.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,8 @@ build:
 	go build -i github.com/kinvolk/habitat-operator/cmd/habitat-operator
 
 linux:
-	env GOOS=linux go build github.com/kinvolk/habitat-operator/cmd/habitat-operator
+	# Compile statically linked binary for linux.
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -i --ldflags="-s" -o habitat-operator github.com/kinvolk/habitat-operator/cmd/habitat-operator
 
 image: linux
 	docker build -t "$(IMAGE):$(TAG)" .
--- a/test/e2e/operator_test.go
+++ b/test/e2e/operator_test.go
@@ -19,11 +19,16 @@ import (
 	"io/ioutil"
 	"net/http"
 	"testing"
+	"time"
 
 	crv1 "github.com/kinvolk/habitat-operator/pkg/habitat/apis/cr/v1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apiv1 "k8s.io/client-go/pkg/api/v1"
+)
+
+const (
+	waitForPorts = 1 * time.Minute
 )
 
 // TestServiceGroupCreate tests service group creation.
@@ -108,6 +113,7 @@ func TestServiceGroupInitialConfig(t *testing.T) {
 	if err := framework.WaitForEndpoints(sgName); err != nil {
 		t.Fatal(err)
 	}
+	time.Sleep(waitForPorts)
 
 	// Get response from Habitat Service.
 	url := fmt.Sprintf("http://%s:30003/", framework.ExternalIP)
@@ -130,6 +136,11 @@ func TestServiceGroupInitialConfig(t *testing.T) {
 	actualMsg := string(bodyBytes)
 	if msg != actualMsg {
 		t.Fatalf("Initial Configuration failed. Msg did not match the one expected. Expected: %s got: %s", msg, actualMsg)
+	}
+
+	// Delete Service so it doesn't interfere with other tests.
+	if err := framework.DeleteService(sgName); err != nil {
+		t.Fatal(err)
 	}
 }
 
@@ -176,6 +187,7 @@ func TestServiceGroupFunctioning(t *testing.T) {
 	if err := framework.WaitForEndpoints(sgName); err != nil {
 		t.Fatal(err)
 	}
+	time.Sleep(waitForPorts)
 
 	// Get response from Habitat Service.
 	url := fmt.Sprintf("http://%s:30002/", framework.ExternalIP)
@@ -199,6 +211,11 @@ func TestServiceGroupFunctioning(t *testing.T) {
 	actualMsg := string(bodyBytes)
 	if expectedMsg != actualMsg {
 		t.Fatalf("Habitat Service msg does not match one in default.toml. Expected: %s got: %s", expectedMsg, actualMsg)
+	}
+
+	// Delete Service so it doesn't interfere with other tests.
+	if err := framework.DeleteService(sgName); err != nil {
+		t.Fatal(err)
 	}
 }
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,12 @@
 language: go
 
-# skip installation of dependencies since they are already in vendor/
+sudo: required
+
+# This moves Kubernetes specific config files.
+env:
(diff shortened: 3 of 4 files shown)


LiliC:
FYI: We do have one test which is flaky, one every 10 times maybe, but this is due to Habitat itself, rather then the operator. The problem being that Supervisor runs into a problem and keeps retrying, so our test fail because the pod is already setup and ready to be tested, but application itself is not up and running.

I will open an issue for this random failure on Habitat, as we discussed on slack. Also there was a suggestion to maybe replace the test image with a Rails application example, currently we use the Node.js application. Will open issue for that as well.

Makefile
```diff
@@ -9,7 +9,7 @@ build:
 	go build -i github.com/kinvolk/habitat-operator/cmd/habitat-operator
 
 linux:
-	env GOOS=linux go build github.com/kinvolk/habitat-operator/cmd/habitat-operator
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -i --ldflags="-s" -o habitat-operator github.com/kinvolk/habitat-operator/cmd/habitat-operator
```
asymmetric: Why are all of these changes necessary?

test/e2e/operator_test.go
```diff
@@ -131,6 +133,11 @@ func TestServiceGroupInitialConfig(t *testing.T) {
 	if msg != actualMsg {
 		t.Fatalf("Initial Configuration failed. Msg did not match the one expected. Expected: %s got: %s", msg, actualMsg)
 	}
+
+	// Delete Service so it doesn't interfere with other tests.
```
asymmetric: How can it interfere?

travis-setup.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
asymmetric: I wonder if we could find a better place for this. Maybe under `test/`?

travis-setup.sh
```diff
@@ -0,0 +1,21 @@
...
+# Setup for Travis CI.
+
+# Download kubectl, which is a requirement for using minikube.
+curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.7.0/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+
+# Download minikube.
+curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
+
+echo "starting minikube..."
+sudo minikube start --vm-driver=none --kubernetes-version=v1.7.0
```
asymmetric: Is `sudo` required here?

test/e2e/operator_test.go
```diff
@@ -108,6 +109,7 @@ func TestServiceGroupInitialConfig(t *testing.T) {
 	if err := framework.WaitForEndpoints(sgName); err != nil {
 		t.Fatal(err)
 	}
+	time.Sleep(1 * time.Minute)
```
asymmetric: Can we make a constant out of tihs?

travis-setup.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+# Setup for Travis CI.
+
+# Download kubectl, which is a requirement for using minikube.
+curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.7.0/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
```
asymmetric: Can we extract the kube version to a variable, here and below?

travis-setup.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# Fix the kubectl context, as it's often stale.
+minikube update-context
+
+# Hack for waiting for minikube to be ready.
+echo "sleeping..."
+sleep 3m
+
+echo "creating image"
+make TAG=testing image
```
asymmetric: Idea for a next iteration: use [build stages](https://docs.travis-ci.com/user/build-stages/) to build the image in parallel, as the tests are being run.

travis-setup.sh
LiliC: Will test, it was needed when the config files were under root, but now that is moved to `USER`, we might not need it anymore.

travis-setup.sh
LiliC: Sure we can either do that, or if you prefer we can have it all in the `.travis.yml` file? I just separated as it was getting long and not that readable, but can move it back there to keep it all in one file. WDYT?

test/e2e/operator_test.go
LiliC: Another service is created again in the other test below, so we just want to make sure to remove this service before creating the other one.

Makefile
LiliC: It was not building the operator on Travis CI properly with the previous command, this seems to do the trick.

travis-setup.sh
LiliC: Cool, feel free to open issue for it, so we don't forget. Agreed also about bringing down the time of the tests, we can maybe make the issue generic `Improve tests performance` or something and note that as well.

travis-setup.sh
LiliC: We still need sudo:
```
E0906 13:50:32.697347    4238 start.go:158] Error updating cluster:  error creating file at /usr/local/bin/localkube: open /usr/local/bin/localkube: permission denied
```

LiliC:
@asymmetric think I addressed/answered all the comments and tests are 🍏 . Can you PTALA, thanks!

travis-setup.sh
asymmetric: Is that because we move the binaries to `/usr/local/bin` though?

travis-setup.sh
asymmetric: Yeah I think it could be a good idea to move it back there, as it's only used in the context of Travis.

Makefile
asymmetric: I see. Is there any documentation we can point to to explain why they're there?

asymmetric:
Addressed comments, including one that's probably hidden, so make sure to expand ;)

Makefile
LiliC: Which ones do you mean? The `go build` flags, are well documented already. And the variables think those are self-explanatory. But I think both are documented enough, not sure we need to leave a comment. :)

travis-setup.sh
LiliC: Got it from the minikube installation [instructions](https://github.com/kubernetes/minikube#linux). Not sure it matters if we do `sudo` or not, as it's just for tests setup.

LiliC:
Done @asymmetric. Can you have another look, danke!

Makefile
asymmetric: If I would look at the Makefile, I wouldn't know why they work, and it seems they have been added "heuristically". So I think it would make sense to justify their presence.

TL;DR: If they're needed, we should explain why IMO.

travis-setup.sh
asymmetric: True, it doesn't really matter, but it's more hygienic IMO to only require privileges when they're strictly needed, as a general rule.

LiliC: Good point, will try to get rid of it.

PR #92, opened 2017-09-15
always add peer-watch-file flag

LiliC:
So far we were only passing the flag when we have `leader` topology. As agreed in https://github.com/kinvolk/habitat-operator/issues/49#issuecomment-325930580, we need to pass this even for standalone topologies.

As for the other change, I came across this error while working on the `--bind` and needed some time to figure out what was wrong as it only stated `too few instances`. This also fixes the formatting, as  `spec.Count` is a number.

### DIFF
--- a/pkg/habitat/controller/utils.go
+++ b/pkg/habitat/controller/utils.go
@@ -42,7 +42,7 @@ func validateCustomObject(sg crv1.ServiceGroup) error {
 	case crv1.TopologyStandalone:
 	case crv1.TopologyLeader:
 		if spec.Count < leaderFollowerTopologyMinCount {
-			return fmt.Errorf("too few instances: %s", spec.Count)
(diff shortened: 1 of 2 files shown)


pkg/habitat/controller/utils.go
```diff
@@ -42,7 +42,7 @@ func validateCustomObject(sg crv1.ServiceGroup) error {
 	case crv1.TopologyStandalone:
 	case crv1.TopologyLeader:
 		if spec.Count < leaderFollowerTopologyMinCount {
-			return fmt.Errorf("too few instances: %s", spec.Count)
+			return fmt.Errorf("too few instances: %d, leader topology requires at least %d", spec.Count, leaderFollowerTopologyMinCount)
```
asymmetric: This is called "leader-follower" in the [hab docs](https://www.habitat.sh/docs/run-packages-topologies/), so I'd change it here.

Elsewhere we use `leader` but that's for code, here it's written English so I'd go with the proper name.

PR #93, opened 2017-09-19
form ring that spans across SG's

LiliC:
As agreed, we need this for the `--bind` feature. This implements only one ring that spans even when we have multiple ServiceGroups.

Owner reference in the ConfigMap was removed, because we do not want to bind the ConfigMap to the SG, because when it gets deleted so does the ConfigMap.

Named the ConfigMap `peer-watch-file`, but feel free to suggest another name.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -45,8 +45,9 @@ const (
 	userTomlFile = "user.toml"
 	configMapDir = "/habitat-operator"
 
-	peerFilename = "peer-ip"
-	peerFile     = "peer-watch-file"
+	peerFilename  = "peer-ip"
+	peerFile      = "peer-watch-file"
+	configMapName = peerFile
 
 	// The key under which the ring key is stored in the Kubernetes Secret.
 	ringSecretKey = "ring-key"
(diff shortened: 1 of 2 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -490,7 +490,7 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
 							VolumeSource: apiv1.VolumeSource{
 								ConfigMap: &apiv1.ConfigMapVolumeSource{
 									LocalObjectReference: apiv1.LocalObjectReference{
-										Name: configMapName(sg.Name),
+										Name: peerFile,
```
asymmetric: Only problem with this is the user won't be able to have more than one ring running in the cluster, cause otherwise the names would clash.

It's not going to happen much, but there's also no real benefit in renaming, so I'd leave it as is (or choose another name that changes with the SG).

pkg/habitat/controller/controller.go
```diff
@@ -697,15 +697,7 @@ func (hc *HabitatController) getServiceGroupFromPod(pod *apiv1.Pod) (*crv1.Servi
...
-			Name: configMapName(sgName),
-			OwnerReferences: []metav1.OwnerReference{
-				{
-					APIVersion: "extensions/v1beta1",
-					Kind:       "Deployment",
-					Name:       sgName,
-					UID:        parentUID,
-				},
-			},
+			Name: peerFile,
```
asymmetric: If you decide to keep the rename, I think it would make sense to have a `var configMapName = peerFile` at the top of the file, so that this would be `Name: configMapName` instead.

pkg/habitat/controller/controller.go
LiliC: Yes it's what we agreed on, to only have one ring, otherwise we cannot bind two services together. The feedback from Habitat was also they prefer one ring.

Issue #53, opened 2017-08-15
Add support for binding services


asymmetric:
More info [here](https://www.habitat.sh/docs/run-packages-binding/).

Can we do an initial implementation, only supporting the `--bind` startup flag, without updates?

This depends on https://github.com/habitat-sh/habitat/issues/2735.

PR #94, opened 2017-09-20
add support for bind

LiliC:
This PR introduces [binding](https://www.habitat.sh/docs/run-packages-binding/). Bind is when one service connects to another forming a producer/consumer relationship. The image taken is a simple [Go server](https://github.com/kinvolk/habitat-server-go-example) that displays the port the bound database listens on.

I also introduced tags for the images we use in tests, this should prevent any accidental modifying.

This closes https://github.com/kinvolk/habitat-operator/issues/53

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,10 @@ clean-test:
 	kubectl delete sg mytutorialapp
 	kubectl delete sg test-service-group
 	kubectl delete sg test-standalone
+	kubectl delete sg test-bind-go
+	kubectl delete sg test-bind-db
 	kubectl delete crd servicegroups.habitat.sh
 	kubectl delete pod habitat-operator
 	kubectl delete secret mytutorialapp
-	kubectl delete service mytutorialapp
-	kubectl delete service test-service-group
 
 .PHONY: build test linux image e2e clean-test
--- a/examples/bind/README.md
+++ b/examples/bind/README.md
@@ -1,0 +1,13 @@
+# Runtime binding
+
+This demonstrates how to run two Habitat Services with a [binding](https://www.habitat.sh/docs/run-packages-binding/) between them.
+
+After the Habitat Operator is up and running, execute the following command from the root of this repository:
+
+```
+kubectl create -f examples/bind/service_group.yml
+```
+
+This will deploy two `ServiceGroup`s, a simple HTTP server written in Go that will be bound to a PostgreSQL database. The Go server will display the port number the database listens on.
+
+When running on minikube, it can be accessed under port `5555` of the minikube VM. `minikube ip` can be used to retrieve the IP.
--- a/examples/bind/service_group.yml
+++ b/examples/bind/service_group.yml
@@ -1,0 +1,40 @@
+apiVersion: habitat.sh/v1
+kind: ServiceGroup
+metadata:
+  name: db
+spec:
+  image: kinvolk/postgresql-hab
+  count: 1
+  habitat:
+    topology: standalone
+---
+apiVersion: habitat.sh/v1
+kind: ServiceGroup
+metadata:
+  name: go
+spec:
+  image: kinvolk/bindgo-hab
+  count: 1
+  habitat:
+    topology: standalone
+    bind:
+      # Name is the name of the bind specified in the Habitat configuration files.
+      - name: db
+        # Service is the name of the service this bind refers to.
+        service: postgresql
+        # Group is the group of the service this bind refers to.
+        group: default
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: go-service
+spec:
+  selector:
+    service-group: go
+  type: NodePort
+  ports:
+  - name: web
+    nodePort: 30001
+    port: 5555
+    protocol: TCP
--- a/pkg/habitat/apis/cr/v1/types.go
+++ b/pkg/habitat/apis/cr/v1/types.go
@@ -62,6 +62,18 @@ type Habitat struct {
 	// The name of the secret that contains the ring key.
 	// Optional.
 	RingSecretName string `json:"ringSecretName,omitempty"`
+	// Bind is when one service connects to another forming a producer/consumer relationship.
+	// Optional.
+	Bind []Bind `json:"bind,omitempty"`
+}
+
+type Bind struct {
+	// Name is the name of the bind specified in the Habitat configuration files.
+	Name string `json:"name"`
+	// Service is the name of the service this bind refers to.
+	Service string `json:"service"`
+	// Group is the group of the service this bind refers to.
+	Group string `json:"group"`
 }
 
 type Topology string
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -455,6 +455,15 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
 		"--peer-watch-file", path,
 	)
 
+	// Runtime binding.
+	// One Service connects to another forming a producer/consumer relationship.
+	for _, bind := range sg.Spec.Habitat.Bind {
+		// Pass --bind flag.
+		bindArg := fmt.Sprintf("%s:%s.%s", bind.Name, bind.Service, bind.Group)
+		habArgs = append(habArgs,
+			"--bind", bindArg)
+	}
+
 	base := &appsv1beta1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: sg.Name,
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"reflect"
 	"regexp"
 	"time"
 
@@ -660,12 +661,12 @@ func (hc *HabitatController) syncServiceGroup(key string) error {
 }
 
 func (hc *HabitatController) serviceGroupNeedsUpdate(oldSG, newSG *crv1.ServiceGroup) bool {
-	if oldSG.Spec != newSG.Spec {
+	if reflect.DeepEqual(oldSG.Spec, newSG.Spec) {
 		level.Debug(hc.logger).Log("msg", "Update ignored as it didn't change ServiceGroup spec", "sg", newSG)
-		return true
+		return false
 	}
 
-	return false
+	return true
 }
 
 func (hc *HabitatController) podNeedsUpdate(oldPod, newPod *apiv1.Pod) bool {
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -29,7 +29,6 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
 	apiv1 "k8s.io/client-go/pkg/api/v1"
 	appsv1beta1 "k8s.io/client-go/pkg/apis/apps/v1beta1"
@@ -204,7 +203,7 @@ func (hc *HabitatController) handleServiceGroupCreation(sg *crv1.ServiceGroup) e
 	}
 
 	// Handle creation/updating of peer IP ConfigMap.
-	if err := hc.handleConfigMap(sg, d.UID); err != nil {
+	if err := hc.handleConfigMap(sg); err != nil {
 		return err
 	}
 
@@ -242,7 +241,7 @@ func (hc *HabitatController) writeLeaderIP(cm *apiv1.ConfigMap, ip string) error
 	return nil
 }
 
-func (hc *HabitatController) handleConfigMap(sg *crv1.ServiceGroup, deploymentUID types.UID) error {
+func (hc *HabitatController) handleConfigMap(sg *crv1.ServiceGroup) error {
 	runningPods, err := hc.getRunningPods(sg.Namespace)
 	if err != nil {
 		return err
@@ -250,7 +249,7 @@ func (hc *HabitatController) handleConfigMap(sg *crv1.ServiceGroup, deploymentUI
 
 	if len(runningPods) == 0 {
 		// No running Pods, create an empty ConfigMap.
-		newCM := newConfigMap(sg.Name, deploymentUID, "")
+		newCM := newConfigMap(sg.Name, "")
 
 		cm, err := hc.config.KubernetesClientset.CoreV1().ConfigMaps(sg.Namespace).Create(newCM)
 		if err != nil {
@@ -279,7 +278,7 @@ func (hc *HabitatController) handleConfigMap(sg *crv1.ServiceGroup, deploymentUI
 	// There are running Pods, add the IP of one of them to the ConfigMap.
 	leaderIP := runningPods[0].Status.PodIP
 
-	newCM := newConfigMap(sg.Name, deploymentUID, leaderIP)
+	newCM := newConfigMap(sg.Name, leaderIP)
 
 	cm, err := hc.config.KubernetesClientset.CoreV1Client.ConfigMaps(sg.Namespace).Create(newCM)
 	if err != nil {
@@ -705,7 +704,7 @@ func (hc *HabitatController) getServiceGroupFromPod(pod *apiv1.Pod) (*crv1.Servi
 	return sg, nil
 }
 
-func newConfigMap(sgName string, parentUID types.UID, ip string) *apiv1.ConfigMap {
+func newConfigMap(sgName string, ip string) *apiv1.ConfigMap {
 	return &apiv1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: configMapName,
--- a/test/e2e/framework/helpers.go
+++ b/test/e2e/framework/helpers.go
@@ -27,13 +27,13 @@ import (
 )
 
 // NewStandaloneSG returns a new Standalone ServiceGroup.
-func (f *Framework) NewStandaloneSG(sgName, group string, secret bool) *crv1.ServiceGroup {
-	sg := crv1.ServiceGroup{
+func NewStandaloneSG(sgName, group, image string) *crv1.ServiceGroup {
+	return &crv1.ServiceGroup{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: sgName,
 		},
 		Spec: crv1.ServiceGroupSpec{
-			Image: "kinvolk/nodejs-hab",
+			Image: image,
 			Count: 1,
 			Habitat: crv1.Habitat{
 				Group:    group,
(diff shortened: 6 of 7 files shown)


Makefile
```diff
@@ -26,10 +26,10 @@ clean-test:
 	kubectl delete sg mytutorialapp
 	kubectl delete sg test-service-group
 	kubectl delete sg test-standalone
+	kubectl delete sg test-bind-go
+	kubectl delete sg test-bind-db
```
asymmetric: This is getting a bit long :)

Can we use a namespace instead?

examples/bind/README.md
```diff
@@ -0,0 +1,13 @@
+# Runtime binding
+
+This demonstrates how to run two Habitat Services with [binding](https://www.habitat.sh/docs/run-packages-binding/).
```
asymmetric: I would say "how to bind two Habitat Services together".

examples/bind/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Run the following command from the root of this repository:
+
+```
+kubectl create -f examples/bind/service_group.yml
+```
+
+This will deploy two `ServiceGroup`s, a simple Go server that will be bound to PostgreSQL database. The Go server will display the port number of the database.
```
asymmetric: * I would explain that it's an HTTP server, as I find "Go server" by itself isn't clear enough.
* bound to *a*
* the port number the database listens on

examples/bind/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+When running on minikube, it can be accessed under port `5555`, of the minikube VM. `minikube ip` can be used to retrieve the IP.
```
asymmetric: Extra comma after `5555`.

pkg/habitat/apis/cr/v1/types.go
```diff
@@ -62,6 +62,17 @@ type Habitat struct {
 	// The name of the secret that contains the ring key.
 	// Optional.
 	RingSecretName string `json:"ringSecretName,omitempty"`
+	// Bind is when one service connects to another forming a producer/consumer relationship.
+	Bind Bind `json:"bind"`
+}
+
+type Bind struct {
+	// Name is the name of the bind specified in Habitat configuration files (plan.sh).
```
asymmetric: *the* Habitat. Also, is it multiple files or just `plan.sh`?

examples/bind/service_group.yml
```diff
@@ -0,0 +1,37 @@
...
+spec:
+  image: kinvolk/postgresql-hab
+  count: 1
+  habitat:
+    topology: standalone
+---
+apiVersion: habitat.sh/v1
+kind: ServiceGroup
+metadata:
+  name: go
```
asymmetric: Where thre some constraints on this key, or other keys? If so, they should probably be mentioned either here or in the README (or both).

pkg/habitat/apis/cr/v1/types.go
```diff
@@ -62,6 +62,17 @@ type Habitat struct {
 	// The name of the secret that contains the ring key.
 	// Optional.
 	RingSecretName string `json:"ringSecretName,omitempty"`
+	// Bind is when one service connects to another forming a producer/consumer relationship.
```
asymmetric: Could you add `Optional` on a newline?

examples/bind/README.md
krnowak: with a binding between them.

examples/bind/README.md
krnowak: Does this assume that the habitat operator is running? If so, maybe mention it here together with the link to docs about how to run it?

pkg/habitat/apis/cr/v1/types.go
```diff
@@ -62,6 +62,17 @@ type Habitat struct {
...
+	Bind Bind `json:"bind"`
```
asymmetric: Not sure whether we want to have `omitempty` on optional fields, but we should be consistent.

pkg/habitat/controller/controller.go
```diff
@@ -455,6 +454,15 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
 		"--peer-watch-file", path,
 	)
 
+	// Runtime binding.
+	// One Service connects to another forming a producer/consumer relationship.
+	if sg.Spec.Habitat.Bind.Name != "" {
+		// Pass --bind flag.
+		bindArg := fmt.Sprintf("%s:%s.%s", sg.Spec.Habitat.Bind.Name, sg.Spec.Habitat.Bind.Service, sg.Spec.Habitat.Bind.Group)
```
asymmetric: This line might look a bit prettier if we extracted `sg.Spec.Habitat.Bind` to a variable beforehand. My 2c.

pkg/habitat/controller/controller.go
```diff
@@ -695,7 +703,7 @@ func (hc *HabitatController) getServiceGroupFromPod(pod *apiv1.Pod) (*crv1.Servi
 	return sg, nil
 }
 
-func newConfigMap(sgName string, parentUID types.UID, ip string) *apiv1.ConfigMap {
+func newConfigMap(sgName string, ip string) *apiv1.ConfigMap {
```
asymmetric: I'm fine when we do some boy/girlscouting, but it should be mentioned in the PR description (or in an inline comment like this one), otherwise it's hard to see how it relates to the PR.

test/e2e/framework/helpers.go
```diff
@@ -27,24 +27,35 @@ import (
...
 				Group:    group,
 				Topology: crv1.TopologyStandalone,
 			},
 		},
 	}
+}
+
+// SGWithConfig appends config fields to the ServiceGroup.
+func (f *Framework) SGWithConfig(sg crv1.ServiceGroup) *crv1.ServiceGroup {
+	sg.Spec.Habitat.Config = sg.ObjectMeta.Name
```
asymmetric: I don't understand what's going on here. Is the secret name the same as the SG name?

I think we should rename the `Config` field by the way. It's a name,whereas intuitively I'd think it's an actual config. So either `ConfigName`, or `Config.Name` would work I think.

That's why lines like the above are kinda hard to grok I think.

test/e2e/operator_test.go
```diff
@@ -257,3 +259,92 @@ func TestServiceGroupDelete(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+// TestBindFunctioning tests that operator correctly created two Habitat Services and bound them together.
```
asymmetric: *the* operator

test/e2e/operator_test.go
```diff
@@ -257,3 +259,92 @@ func TestServiceGroupDelete(t *testing.T) {
...
+func TestBindFunctioning(t *testing.T) {
+	//Create two SG to test binding between them.
```
asymmetric: Space after `//`. OCD!!

test/e2e/operator_test.go
```diff
@@ -257,3 +259,92 @@ func TestServiceGroupDelete(t *testing.T) {
...
+
+	bodyBytes, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// This msg is set in the config of the kinvolk/bindgo-hab Go Habitat Service.
+	expectedMsg := "hello from port: 5432"
+	actualMsg := string(bodyBytes)
+	if !strings.Contains(actualMsg, expectedMsg) {
```
asymmetric: Would `!==` also work?

test/e2e/operator_test.go
```diff
@@ -257,3 +259,92 @@ func TestServiceGroupDelete(t *testing.T) {
...
+		t.Fatalf("Habitat Service msg does not match one in default.toml. Expected: *%s* got: *%s*", expectedMsg, actualMsg)
+	}
+
+	// Delete Service so it doesn't interfere with other tests.
```
asymmetric: Shouldn't we delete everything after each test?

asymmetric:
On the commit messages part: I agree, but we have also squashed commits in the past, so the ones you see in the PR are not necessarily in their final form.

test/e2e/operator_test.go
LiliC: Only the Services interfere with each other if test run is out of order. Otherwise no need to delete everything, when we do it locally we run `make clean` when the tests are finished.

pkg/habitat/controller/controller.go
LiliC: Will do next time, I did do a separate commit for it, but can see how it can be confusing.

krnowak:
@asymmetric: We were usually squashing commits that addressed the review issues. But in general we should avoid lumping everything into one big commit.

pkg/habitat/apis/cr/v1/types.go
LiliC: It is multiple `plan.sh` files, also you then use that name in other configuration files where you want to actually get the value of the bind. I will omit `plan.sh` to avoid confusion.

Makefile
asymmetric: Added an issue for it btw, #96.

LiliC: Thanks!

test/e2e/operator_test.go
LiliC: Not sure I know of `!==`. Can you elaborate?

asymmetric: Sorry, meant `!=`. Was just wondering if using `Contains` is necessary whether just comparing the two strings straight away.

LiliC: Tried that, but `actualMsg` can contain whitespace and newlines or different formatting. The only thing we care about is that it contains that string, so I went with that route rather then getting rid of whitespace. But am open for suggestions.

examples/bind/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This demonstrates how to run two Habitat Services with a [binding](https://www.habitat.sh/docs/run-packages-binding/) between them.
+
+After the Habitat Operator is up and running, execute the following command from the root of this repository:
+
+```
+kubectl create -f examples/bind/service_group.yml
+```
+
+This will deploy two `ServiceGroup`s, a simple HTTP server written in Go that will be bound to a PostgreSQL database. The Go server will display the port number the dabase listens on.
```
krnowak: `s/dabase/database/`

pkg/habitat/apis/cr/v1/types.go
```diff
@@ -62,6 +62,18 @@ type Habitat struct {
 	// The name of the secret that contains the ring key.
 	// Optional.
 	RingSecretName string `json:"ringSecretName,omitempty"`
+	// Bind is when one service connects to another forming a producer/consumer relationship.
+	// Optional.
+	Bind Bind `json:"bind,omitempty"`
```
krnowak: This should be an array, shouldn't it? You can have a service that binds to several other services.

test/e2e/operator_test.go
```diff
@@ -30,12 +31,14 @@ import (
 const (
 	waitForPorts  = 1 * time.Minute
 	configMapName = "peer-watch-file"
+
+	nodejsImage = "kinvolk/nodejs-hab:test"
```
krnowak: Uh, something does not look right to me. So before this PR, `kinvolk/nodejs-hab` was hardcoded in `NewStandaloneSG` function. Now, you can pass image name to the function, so you are passing a different thing now - `kinvolk/nodejs-hab:test` (note the `:test` suffix).

examples/bind/README.md
asymmetric: I think she was listening to [this](https://www.youtube.com/watch?v=3cYwggM2UGk) while typing.

test/e2e/operator_test.go
LiliC: As discussed offline, this is already mentioned in the description. 
> I also introduced tags for the images we use in tests, this should prevent any accidental modifying.

pkg/habitat/apis/cr/v1/types.go
LiliC: Thanks, wasn't aware you can have multiple binds.

pkg/habitat/controller/controller.go
```diff
@@ -651,11 +662,11 @@ func (hc *HabitatController) syncServiceGroup(key string) error {
 }
 
 func (hc *HabitatController) serviceGroupNeedsUpdate(oldSG, newSG *crv1.ServiceGroup) bool {
-	if oldSG.Spec != newSG.Spec {
-		level.Debug(hc.logger).Log("msg", "Update ignored as it didn't change ServiceGroup spec", "sg", newSG)
+	if !reflect.DeepEqual(oldSG.Spec, newSG.Spec) {
```
LiliC: This had to be done due to changing our struct to include a slice, comparison did not work anymore. Also changed the log, as I believe it was not correct before @asymmetric?

test/e2e/framework/helpers.go
```diff
@@ -27,25 +27,30 @@ import (
 )
 
 // NewStandaloneSG returns a new Standalone ServiceGroup.
-func (f *Framework) NewStandaloneSG(sgName, group string, secret bool) *crv1.ServiceGroup {
-	sg := crv1.ServiceGroup{
+func NewStandaloneSG(sgName, group, image string) *crv1.ServiceGroup {
```
LiliC: Also got rid of this @krnowak, as we discussed offline earlier, but I think you might have forgotten to comment.

LiliC:
Think I addressed all your comments @krnowak @asymmetric, can you have another look. I left some comments why I did some extra changes.

pkg/habitat/controller/controller.go
asymmetric: Good catch!

But I'd reverse the conditional and end with `return true` as that's the positive case.

pkg/habitat/controller/controller.go
```diff
@@ -456,12 +457,13 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
 
 	// Runtime binding.
 	// One Service connects to another forming a producer/consumer relationship.
-	if sg.Spec.Habitat.Bind.Name != "" {
-		// Pass --bind flag.
-		bind := sg.Spec.Habitat.Bind
-		bindArg := fmt.Sprintf("%s:%s.%s", bind.Name, bind.Service, bind.Group)
-		habArgs = append(habArgs,
-			"--bind", bindArg)
+	if sg.Spec.Habitat.Bind != nil {
```
asymmetric: Alternatively, `len(sg.Spec.Habitat.Bind) > 0`

examples/bind/service_group.yml
```diff
@@ -19,11 +19,11 @@ spec:
     topology: standalone
     bind:
       # Name is the name of the bind specified in the Habitat configuration files.
-      name: db
-      # Service is the name of the service this bind refers to.
-      service: postgresql
-      # Group is the group of the service this bind refers to.
-      group: default
+      - name: pb
```
asymmetric: `db` and `postgresql` sounded better. Or `postgres`. :)

LiliC: Yeah was testing something locally and managed to rename it. 😄

pkg/habitat/controller/controller.go
LiliC: null value of a slice is `nil` and that makes it more clear when the value is unset by the user, and the `len(...) > 0` case is already handled by the for loop.

LiliC:
@asymmetric Thanks for your comments, addressed them. Can you have another look.

test/e2e/framework/helpers.go
```diff
@@ -27,25 +27,30 @@ import (
...
-	}
-	return &sg
+// SGWithConfig appends config fields to the ServiceGroup.
+func SGWithConfig(sg *crv1.ServiceGroup) {
+	sg.Spec.Habitat.Config = sg.ObjectMeta.Name
+}
+
+// SGWithBind appends bind fields to the ServiceGroup.
+func SGWithBind(sg *crv1.ServiceGroup, bindName, bindService string) {
+	sg.Spec.Habitat.Bind = []crv1.Bind{crv1.Bind{Name: bindName, Service: bindService, Group: sg.Spec.Habitat.Group}}
```
asymmetric: I'd split this over multiple lines, as like this it's a bit of an eyesore ;)

test/e2e/operator_test.go
```diff
@@ -30,12 +32,14 @@ import (
...
 )
 
 // TestServiceGroupCreate tests service group creation.
 func TestServiceGroupCreate(t *testing.T) {
 	sgName := "test-standalone"
-	sg := framework.NewStandaloneSG(sgName, "foobar", false)
+	sg := utils.NewStandaloneSG(sgName, "foobar", nodejsImage)
```
asymmetric: I like `utils` much better :+1:

pkg/habitat/controller/controller.go
asymmetric: Yeah, you could have just had the loop without conditional, that was my point in suggesting the change. But it's just a stylistic preference.

pkg/habitat/controller/controller.go
```diff
@@ -455,6 +455,17 @@ func (hc *HabitatController) newDeployment(sg *crv1.ServiceGroup) (*appsv1beta1.
 		"--peer-watch-file", path,
 	)
 
+	// Runtime binding.
+	// One Service connects to another forming a producer/consumer relationship.
+	if sg.Spec.Habitat.Bind != nil {
```
krnowak: You can drop this `if` and do `range` directly. Go handles `nil`s gracefully here.

test/e2e/framework/helpers.go
```diff
@@ -27,25 +27,34 @@ import (
...
 		},
 	}
+}
 
-	if secret {
-		sg.Spec.Habitat.Config = sgName
-	}
-	return &sg
+// SGWithConfig appends config fields to the ServiceGroup.
+func SGWithConfig(sg *crv1.ServiceGroup) {
```
krnowak: I'd name it `SetupConfigInSG` or `AddConfigToSG`, but rather the former - for the latter I'd expect to pass some config as a function parameter.

test/e2e/framework/helpers.go
```diff
@@ -27,25 +27,34 @@ import (
...
+	sg.Spec.Habitat.Config = sg.ObjectMeta.Name
+}
+
+// SGWithBind appends bind fields to the ServiceGroup.
+func SGWithBind(sg *crv1.ServiceGroup, bindName, bindService string) {
```
krnowak: I'd name it `AddBindToSG`. And please use `append`.

test/e2e/operator_test.go
krnowak: Maybe add a comment about it?

pkg/habitat/controller/controller.go
LiliC: @asymmetric already mentioned this, but I feel that handling nil explicitly here shows that a user has actually set the Bind on purpose. For me it helps with reading flow of the code. But since you both suggest it, can change of course.

krnowak: I haven't noticed, but please do it, though. User won't likely read the code to figure out whether the bindings are implicit or explicit, and for me handling `nil` did not convey that information.

LiliC: @krnowak So if I understand correctly you want me to add to docs that this is optional, remove the `nil` and instead add a comment in the code as well?

LiliC:
done, can you have another look @krnowak, thanks.

krnowak:
Code LFAD, so you could start on preparing the proper commits now. Thanks. :)

Issue #96, opened 2017-09-21
Use Namespace in tests


asymmetric:
See https://github.com/kinvolk/habitat-operator/pull/94#discussion_r140207694

PR #100, opened 2017-09-26
test: use testing namespace in e2e tests

LiliC:
All the resources in the tests are created in the "testing" Kubernetes namespace.
This makes it easier to run the tests locally as they do not interfere with any existing resources. Also helps with the test cleanup.

This closes https://github.com/kinvolk/habitat-operator/issues/96.

### DIFF
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -28,6 +28,10 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+const (
+	TestNs = "testing"
+)
+
 type Framework struct {
 	Image      string
 	KubeClient kubernetes.Interface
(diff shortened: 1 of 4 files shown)


test/e2e/framework/framework.go
```diff
@@ -28,6 +28,10 @@ import (
 	"k8s.io/client-go/tools/clientcmd"
 )
 
+const (
+	TestNs = "testing"
```
asymmetric: Should probably be `TestNS`

LiliC: Followed the naming as it's `Namespace` in their [docs](https://kubernetes.io/docs/tasks/administer-cluster/namespaces-walkthrough/). But can change if it makes more sense `TestNS` for you? Or to make it even more readable we can rename it to `TestNamespace`?

asymmetric: Ah, good point. Ns then.

PR #104, opened 2017-10-05
Fix user.toml example

asymmetric:
The serialization was incorrect, so the secret was never mounted.

Also adds `omitempty` since it's an optional key.

### DIFF
--- a/examples/config/service_group.yml
+++ b/examples/config/service_group.yml
@@ -32,10 +32,12 @@ spec:
     service-group: mytutorialapp
   type: NodePort
   ports:
+  # This endpoint displays the message from the secret
   - name: web
     nodePort: 30001
     port: 5555
     protocol: TCP
+  # This endpoint exposes the Habitat supervisor API
   - name: http-gateway
     nodePort : 32767
     port: 9631
--- a/pkg/habitat/apis/cr/v1/types.go
+++ b/pkg/habitat/apis/cr/v1/types.go
@@ -58,7 +58,7 @@ type Habitat struct {
 	// ConfigSecretName is the name of the Secret containing the config the user has previously created.
 	// The file with this name is mounted inside of the pod. Habitat will
 	// use it for initial configuration of the service.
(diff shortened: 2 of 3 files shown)


pkg/habitat/apis/cr/v1/types.go
```diff
@@ -58,7 +58,7 @@ type Habitat struct {
 	// ConfigSecretName is the name of the Secret containing the config the user has previously created.
 	// The file with this name is mounted inside of the pod. Habitat will
 	// use it for initial configuration of the service.
-	ConfigSecretName string `json:"config"`
+	ConfigSecretName string `json:"configSecretName,omitempty"`
```
LiliC: funny enough I just found this right now when creating the bind+config demo 👍 😄

examples/config/service_group.yml
```diff
@@ -32,10 +32,12 @@ spec:
     service-group: mytutorialapp
   type: NodePort
   ports:
+  # this endpoint is displays the message from the secret
   - name: web
     nodePort: 30001
     port: 5555
     protocol: TCP
+  # this endpoint exposes the habitat supervisor API
```
LiliC: Capital `Habitat`.

examples/config/service_group.yml
```diff
@@ -32,10 +32,12 @@ spec:
     service-group: mytutorialapp
   type: NodePort
   ports:
+  # This endpoint displays the message from the secret
```
zeenix: I don't really know what's going on but these changes seem clearly unrelated to the commit. I know I'm too late and it's merged already, just for future reference. :)

asymmetric: And originally it was a separate commit! I then thought it would be better not to have only a commit adding comments, and squashed it.

zeenix: ah ok. Don't be afraid of trivial small commits. Best to always follow single logical-change-single-patch policy, so reverting is easy for one thing. :)

PR #111, opened 2017-10-09
CHANGELOG: add release notes for 0.1.0

LiliC:
This PR introduces a `CHANGELOG` file. Also added the notes that went to the `0.1.0` release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Habitat operator CHANGELOG
+
+## 0.1.0/2017-10-9
+
+First release of Habitat operator. Highlight of features:
+
+* [FEATURE] Support for creating/managing Habitat Services
+* [FEATURE] Support for different Habitat topologies
+* [FEATURE] Support for bind feature
+* [FEATURE] Support for initial configuration
+* [FEATURE] Support for encrypted ring communication
+


CHANGELOG.md
```diff
@@ -0,0 +1,12 @@
...
+
+## 0.1.0/2017-10-9
+
+First release of Habitat operator. Highlight of features:
+
+* [FEATURE] Support for creating/managing Habitat Services
+* [FEATURE] Support for different Habitat topologies
+* [FEATURE] Support for bind feature
+* [FEATURE] Support for initial configuration
+* [FEATURE] Support for encrypted ring communication
```
zeenix: I'm unable to appreciate redundancy unless it's absolutely required. I guess  "[FEATURE]" tag is a common practice so needs to be there but I don't see the point of "Support for " in the beginning of every feature title/description (it's implied).

LiliC: As discussed offline: We already have this in the Release notes, this is just importing it to the `CHANGELOG` file.

krnowak:
An opinion: we should just close it. If we want to follow Habitat's change log, then you can notice that it does not have an entry for the first released version. Which makes sense, because there is no previous release on top of which the current release has improvements. So let's add the CHANGELOG when we actually release 0.2.0.

LiliC:
@krnowak Agree with that. Closing. 👍

Issue #66, opened 2017-08-21
Add Deployment watcher


asymmetric:
We should have watchers for all resources that affect `Habitat`s.

We have a pod watcher, but we still need to add a Deployment one.

Depends on #63.

Issue #65, opened 2017-08-21
Use cache instead of making API calls


asymmetric:
Whenever possible, we should use the cache returned by the `cache.NewInformer` function, instead of making API calls, to retrieve objects from the API.

Depends on #63.

asymmetric:
After looking at this, I believe it's not possible to use the cache for Pods at the moment, since all listing functions [only accept labels](https://github.com/kubernetes/client-go/blob/9675d048204baf818bb24f7c74d0e64f6254902e/listers/core/v1/pod.go#L31), whereas we also need to [filter by field](https://github.com/kinvolk/habitat-operator/blob/b81771cdb1725afd5f3760a86856256bb9e81a40/pkg/habitat/controller/controller.go#L207-L209).

asymmetric:
Closed by #156.

Opened #180 for the remaining case.

PR #113, opened 2017-10-19
React on events

LiliC:
So far we were only reacting on `Habitat` and `Pod` events. This PR adds `Deployment` and `ConfigMap` events. This PR also stores the cache. This cache will be used in the subsequent PR and it will optimise the unnecessary calls to the API. My reasons for splitting is for easier review.

Couple of points:
- In the `handleHab` function I extracted the handling into separate functions to be consistent with the other resource handlers.
- Adding enqueuing in the `handlePodAdd` improves the time the IP is written into the `peer-watch-file` CM.
- I have explored reacting to `Secrets` as well and storing them. But we do not store any information about which `Secret` belongs to which `Habitat`. Therefor we cannot enqueue the correct `Habitat` object and would need to enqueue al of the existing `Habitat`s in all namespaces, which would just slow things down.
- We do react on `ConfigMap` events even if it's one `CM` for all the `Habitat`s. We do this because if there is an add/update/delete event on the peer-watch-file `CM` we do want to reconcile again asap no matter the `Habitat`.
- As for the `Pod`s as @asymmetric already [discovered](https://github.com/kinvolk/habitat-operator/issues/65#issuecomment-324927718) we cannot use the cache to get rid of the API calls.

Closes https://github.com/kinvolk/habitat-operator/issues/66.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -442,6 +442,9 @@ func (hc *HabitatController) handleConfigMap(h *crv1.Habitat) error {
 			if err := hc.writeLeaderIP(cm, ""); err != nil {
 				return err
 			}
+
+			level.Info(hc.logger).Log("msg", "removed peer IP from ConfigMap", "name", newCM.Name)
+
 			return nil
 		}
 
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -333,10 +333,6 @@ func (hc *HabitatController) handlePodUpdate(oldObj, newObj interface{}) {
 		return
 	}
 
-	if oldPod.ObjectMeta.Labels[crv1.HabitatLabel] != "true" {
-		return
-	}
-
 	newPod, ok2 := newObj.(*apiv1.Pod)
 	if !ok2 {
 		level.Error(hc.logger).Log("msg", "Failed to type assert pod", "obj", newObj)
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -69,8 +69,9 @@ type HabitatController struct {
 	// delay, so that jobs in a crashing loop don't fill the queue.
 	queue workqueue.RateLimitingInterface
 
-	// store is the cache of Habitats retrieved by the ListWatcher.
-	store cache.Store
+	habInformer    cache.SharedIndexInformer
+	deployInformer cache.SharedIndexInformer
+	cMInformer     cache.SharedIndexInformer
 }
 
 type Config struct {
@@ -106,9 +107,14 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
 func (hc *HabitatController) Run(ctx context.Context) error {
 	level.Info(hc.logger).Log("msg", "Watching Habitat objects")
 
-	hc.watchHabitats(ctx)
-
+	hc.cacheHab()
+	hc.cacheDeployment()
+	hc.cacheConfigMap()
 	hc.watchPods(ctx)
+
+	go hc.habInformer.Run(ctx.Done())
+	go hc.deployInformer.Run(ctx.Done())
+	go hc.cMInformer.Run(ctx.Done())
 
 	// Start the synchronous queue consumer.
 	go hc.worker()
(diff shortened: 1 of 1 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -268,6 +448,7 @@ func (hc *HabitatController) handleConfigMap(h *crv1.Habitat) error {
 			if err := hc.writeLeaderIP(cm, ""); err != nil {
 				return err
 			}
+			return nil
```
LiliC: We return early here because we have finished with handling the CM for this event.

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,269 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
-		if !apierrors.IsAlreadyExists(err) {
-			return err
+func (hc *HabitatController) handleDeployAdd(obj interface{}) {
+	d, ok := obj.(*appsv1beta1.Deployment)
+	if !ok {
+		level.Error(hc.logger).Log("msg", "Failed to type assert deployment", "obj", obj)
+		return
+	}
+
+	if d.ObjectMeta.Labels["habitat"] == "true" {
```
krnowak: How about adding a function like `func isHabitatObject(meta *<type of ObjectMeta>) bool` that checks the label?

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,268 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
+	hc.habInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    hc.handleHabAdd,
+		UpdateFunc: hc.handleHabUpdate,
+		DeleteFunc: hc.handleHabDelete,
+	})
+}
+
+func (hc *HabitatController) cacheDeployment() {
+	source := cache.NewListWatchFromClient(
+		hc.config.KubernetesClientset.AppsV1beta1().RESTClient(),
```
krnowak: My lack of client-go knowledge probably leaves me confused about the fact that you pass different things as `cache.Getter` for each watcher. It's `hc.config.HabitatClient` for watching Habitats, `hc.config.KubernetesClientset.AppsV1beta1().RESTClient()` for watching deployments and `hc.config.KubernetesClientset.CoreV1().RESTClient()` for config maps and pods. So for each of those things do you need a separate client or something?

pkg/habitat/controller/controller.go
krnowak: Also, use `crv1.HabitatLabel` instead of `"habitat"`.

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,268 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
+		return
+	}
+
+	if d.ObjectMeta.Labels["habitat"] == "true" {
+		hc.enqueue(obj)
+	}
+}
+
+func (hc *HabitatController) handleCMAdd(obj interface{}) {
+	cm := obj.(*apiv1.ConfigMap)
```
krnowak: In other handlers you check explicitly if the type assert succeeded. Why not doing it here?

krnowak: Also, this code is duplicated in update and delete handlers - maybe just put all this in one function? Unless you have plans in the immediate future to change one of the config map handlers.

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,268 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
+func (hc *HabitatController) handleCMDelete(obj interface{}) {
+	cm := obj.(*apiv1.ConfigMap)
+	if cm.ObjectMeta.Labels["habitat"] == "true" {
+		hc.enqueueNs(cm.GetNamespace())
+	}
+}
+
+func (hc *HabitatController) enqueueNs(ns string) {
+	cache.ListAll(hc.habInformer.GetStore(), labels.Everything(), func(obj interface{}) {
+		h := obj.(*crv1.Habitat)
```
krnowak: Explicit check for type assertion?

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,268 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
+		if h.Namespace == ns {
+			hc.enqueue(h)
 		}
+	})
+}
 
-		d, err = hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(h.Namespace).Get(deployment.Name, metav1.GetOptions{})
+func (hc *HabitatController) handlePodAdd(obj interface{}) {
+	pod := obj.(*apiv1.Pod)
```
krnowak: Explicit type assertion check?

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,268 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
 	}
 
-	return nil
+	newPod, ok2 := newObj.(*apiv1.Pod)
+	if !ok2 {
+		level.Error(hc.logger).Log("msg", "Failed to type assert pod", "obj", newObj)
+		return
+	}
+
+	if !hc.podNeedsUpdate(oldPod, newPod) {
```
krnowak: Shouldn't you check the habitat label on the new pod too before checking if it needs update?

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,268 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
+	}
+
+	h, err := hc.getHabitatFromPod(newPod)
+	if err != nil {
+		if hErr, ok := err.(habitatNotFoundError); !ok {
+			level.Error(hc.logger).Log("msg", hErr)
+			return
+		}
+
+		// This only means the Pod and the Habitat watchers are not in sync.
```
krnowak: Is this harmless?

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,268 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
+	}
+
+	h, err := hc.getHabitatFromPod(pod)
+	if err != nil {
+		if hErr, ok := err.(habitatNotFoundError); !ok {
+			level.Error(hc.logger).Log("msg", hErr)
+			return
+		}
+
+		// This only means the Pod and the Habitat watchers are not in sync.
```
krnowak: Ditto about being harmless.

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,268 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
+	hc.cMInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    hc.handleCMAdd,
+		UpdateFunc: hc.handleCMUpdate,
+		DeleteFunc: hc.handleCMDelete,
+	})
 }
 
-func (hc *HabitatController) handleHabitatCreation(h *crv1.Habitat) error {
-	level.Debug(hc.logger).Log("function", "handleHabitatCreation", "msg", h.ObjectMeta.SelfLink)
+func (hc *HabitatController) watchPods(ctx context.Context) {
```
krnowak: This function is similar to `cacheFoo` functions. Maybe you could make it consistent with the rest? You could:
- either change the `cacheFoo` functions to run the informer in the same function, so you don't need to store them as members of `HabitatController` (with `habInfomer` being an exception, because it seems to be used elsewhere)
- or change the `watchPods` function (and rename it to `cachePod`) to set the to-be-added `podInformer` member of `HabitatController` and run it in the `Run` function (just like you are doing with the `cacheFoo` functions.

pkg/habitat/controller/controller.go
krnowak: We do return nil just a few lines below anyway. Did you add this to avoid the info about config map being created?

pkg/habitat/controller/controller.go
```diff
@@ -656,8 +755,44 @@ func (hc *HabitatController) syncHabitat(key string) error {
...
+	}
+
+	// Create Deployment, if it doesn't already exist.
+	_, err = hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(h.Namespace).Create(deployment)
+	if err != nil {
+		// Was the error due to the Deployment already existing?
+		if !apierrors.IsAlreadyExists(err) {
+			return err
+		}
+		_, err = hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(h.Namespace).Get(deployment.Name, metav1.GetOptions{})
```
krnowak: What's the point in confirming that the deployment exists by trying to get it if the error already stated the the deployment already exists?

pkg/habitat/controller/controller.go
```diff
@@ -709,6 +844,9 @@ func newConfigMap(ip string) *apiv1.ConfigMap {
 	return &apiv1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: configMapName,
+			Labels: map[string]string{
+				crv1.HabitatLabel: "true",
```
krnowak: I'm sometimes wondering if we should make `"true"` also some constant, like `crv1.HabitatLabel` is now. Would it be an overkill?

pkg/habitat/controller/controller.go
LiliC: Yes exactly, otherwise the log msgs were confusing and it stated each time "created peer....". Returning earlier made sense here.

pkg/habitat/controller/controller.go
LiliC: Actually not really. As I mentioned in the description of the PR we do not actually cache the pods, we just react on events. So I thing the name `watchPods` is appropriate.

> ...so you don't need to store them as members of HabitatController

We will use the informer in the very next PR, as mentioned in the description. I split it up for easier review. This is just about reacting on events.

pkg/habitat/controller/controller.go
LiliC: > Also, this code is duplicated in update and delete handlers - maybe just put all this in one function?

Not sure I understand exactly what you mean by that?

pkg/habitat/controller/controller.go
LiliC: I think it's an overkill but I see your point. 👍

pkg/habitat/controller/controller.go
krnowak: The body of the `handleCMAdd`, `handleCMUpdate` and `handleCMDelete` are all the same. So:
- either just make it one function and pass it to the `AddEventHandler` for `hc.cMInformer`,
- or (better IMO) just move this code to a separate function and call it from all the `handleCMFoo` functions.

krnowak: But don't do any of those if the followup PR will make the `handleCMFoo` functions all different.

pkg/habitat/controller/controller.go
krnowak: Maybe add some debug about clearing the config map then?

pkg/habitat/controller/controller.go
krnowak: Alright, let's keep it like that then.

pkg/habitat/controller/controller.go
LiliC: This code was just taken from the already existing codebase. 

But after having a look at the code, I think we do not need to do the check in the first place. As in the `watchPods` function we already filter and react only on events of the pods with the `"habitat" : "true"` label. So will remove the first check as that condition would never happen.
https://github.com/kinvolk/habitat-operator/blob/a388c7738323daaf35126dc09e90bca180031030/pkg/habitat/controller/controller.go#L199

pkg/habitat/controller/controller.go
LiliC: This is just moved from the old codebase.

But I agree with you, and I commented this in the PR that introduced this part of the code. My plan is to actually first do a get from the cache and then create a deployment if it does not exist. I already started implementing this in the part two of this PR, wdyt of that plan?

pkg/habitat/controller/controller.go
LiliC: So each of the rest clients has their own registered resources, so for example the `AppsV1Beta1` knows about the `deployments` resource, etc. 

There are no validations for when we actually do [cache them](https://github.com/kinvolk/habitat-operator/blob/c541b040326c0e01de8d397289cd336c70a97025/vendor/k8s.io/client-go/tools/cache/listwatch.go#L61), so you can pass in any of clients, but if you actually try it out it will print out an error when doing the actual request. The error looks something like this:
```
ERROR: github.com/kinvolk/habitat-operator/pkg/habitat/controller/controller.go:116: Failed to list *v1beta1.Deployment: the server could not find the requested resource
```

tl;dr: We need to pass exactly those clients.

pkg/habitat/controller/controller.go
LiliC: Makes sense, extracted and moved the content of the `enqueueNS` into the `enqueueCM` function.

pkg/habitat/controller/controller.go
LiliC: Will do, although sometimes I feel like we log too much maybe. I personally only look at the logs that actually write the IP into the CM rather then the other logs, ie the ones that contain the IP.

pkg/habitat/controller/controller.go
LiliC: This part of the code was just moved to a more sensible place, the comment was added in a previous PR not by me. But in general syncing happens on each resource event so they will get in the correct state at another point.

LiliC:
@krnowak Think I addressed or answered all the comments you had, can you PTALA.

pkg/habitat/controller/controller.go
krnowak: Alright, thanks.

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,262 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
+
+func (hc *HabitatController) handlePodDelete(obj interface{}) {
+	pod, ok := obj.(*apiv1.Pod)
+	if !ok {
+		level.Error(hc.logger).Log("msg", "Failed to type assert pod", "obj", obj)
+		return
 	}
 
-	return nil
+	if isHabitatObject(&pod.ObjectMeta) == false {
```
krnowak: `if !isHabitatObject(…)`.

pkg/habitat/controller/controller.go
```diff
@@ -623,7 +718,7 @@ func (hc *HabitatController) processNextItem() bool {
 	}
 	defer hc.queue.Done(key)
 
-	err := hc.syncHabitat(key.(string))
+	err := hc.conform(key.(string))
```
krnowak: Maybe also add a comment about this panicking type assertion?

pkg/habitat/controller/controller.go
```diff
@@ -723,3 +856,11 @@ func habitatKeyFromPod(pod *apiv1.Pod) string {
 
 	return key
 }
+
+func isHabitatObject(objMeta *metav1.ObjectMeta) bool {
+	if objMeta.Labels[crv1.HabitatLabel] != "true" {
```
krnowak: Just `return objMeta.Labels[crv1.HabitatLabel] == "true"` should be fine.

pkg/habitat/controller/controller.go
krnowak: Not sure if there is such a thing like "too many logs", but rather "misleading/wrong logs". But sure, more logging can be added when a need arises.

pkg/habitat/controller/controller.go
krnowak: Sure, sounds fine.

pkg/habitat/controller/controller.go
```diff
@@ -120,94 +126,262 @@ func (hc *HabitatController) Run(ctx context.Context) error {
...
+func (hc *HabitatController) enqueueCM(obj interface{}) {
+	cm, ok := obj.(*apiv1.ConfigMap)
+	if !ok {
+		level.Error(hc.logger).Log("msg", "Failed to type assert ConfigMap", "obj", obj)
+		return
+	}
+
+	if isHabitatObject(&cm.ObjectMeta) == true {
+		cache.ListAll(hc.habInformer.GetStore(), labels.Everything(), func(obj interface{}) {
+			h := obj.(*crv1.Habitat)
```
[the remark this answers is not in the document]
LiliC: Actually I just did a check and put it in code rather then a comment.

pkg/habitat/controller/controller.go
LiliC: Added a check here as well.

LiliC:
@krnowak Addressed your suggestions, can you have look again, thanks!

pkg/habitat/controller/controller.go
```diff
@@ -268,6 +449,8 @@ func (hc *HabitatController) handleConfigMap(h *crv1.Habitat) error {
 			if err := hc.writeLeaderIP(cm, ""); err != nil {
 				return err
 			}
+			return nil
+			level.Info(hc.logger).Log("msg", "removed peer IP from ConfigMap", "name", newCM.Name)
```
krnowak: This is dead code. It should be moved above `return nil`.

LiliC: Oh god, wish there was a facepalm emoji on here. :/ Fixed.

krnowak:
LFAD.

PR #114, opened 2017-10-19
docs

LiliC:
Add some note about where to find release images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,8 @@ Start the Habitat operator by running:
 
 ### Running inside a Kubernetes cluster
 
+#### Building image from source
+
 First build the image:
 
     make image
@@ -39,6 +41,10 @@ First build the image:
 This will produce a `kinvolk/habitat-operator` image, which can then be deployed to your cluster.
 
 The name of the generated docker image can be changed with an `IMAGE` variable, for example `make image IMAGE=mycorp/my-habitat-operator`. If the `habitat-operator` name is fine, then a `REPO` variable can be used like `make image REPO=mycorp` to generate the `mycorp/habitat-operator` image. Use the `TAG` variable to change the tag to something else (the default value is taken from `git describe --tags --always`) and a `HUB` variable to avoid using the default docker hub.
+
+#### Using release image
+
+Habitat operator images are located [here](https://hub.docker.com/r/kinvolk/habitat-operator/), they are tagged with the release version.
 
 ### Deploying an example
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -32,6 +32,8 @@ Start the Habitat operator by running:
 
 ### Running inside a Kubernetes cluster
 
+### Building image from source
```
krnowak: This makes the "Running inside a Kubernetes cluster" section empty. Maybe you wanted to use four hashes for the "Building image from source" section?

README.md
```diff
@@ -40,6 +42,10 @@ This will produce a `kinvolk/habitat-operator` image, which can then be deployed
 
 The name of the generated docker image can be changed with an `IMAGE` variable, for example `make image IMAGE=mycorp/my-habitat-operator`. If the `habitat-operator` name is fine, then a `REPO` variable can be used like `make image REPO=mycorp` to generate the `mycorp/habitat-operator` image. Use the `TAG` variable to change the tag to something else (the default value is taken from `git describe --tags --always`) and a `HUB` variable to avoid using the default docker hub.
 
+### Using release image
```
krnowak: Four hashes?

LiliC:
@krnowak good point, wdyt if we change the `Topology` scheme to string, as it is a string in the end?

krnowak:
I'd say that Topology is more like an enumeration, but you already document the possible values. So yeah, string sounds fine. A problem that I see is that we allow either "standalone" or "leader" topologies, but there is also "none" in Habitat. I need to check what is the difference between the "standalone" topology and "none".

LiliC:
@krnowak IIRC topology in Habitat itself always defaults to standalone.

LiliC:
@krnowak changed to `string`. PTAL.

tashimi:
Topology defaults to none, @fnichol can weigh in too, the difference between standalone and none is that none will never update itself

LiliC:
@tashimi Interesting, as when I was reading about topologies, I came across [this](https://www.habitat.sh/docs/glossary/#glossary-services) in the docs at some point so I relied on that info.
> Standalone
>This is the default topology, useful for services inside a group that are completely independent from one another. 

If that's the case we will make the adjustment in the operator. Thanks for this!

tashimi:
I meant to update those docs when I was going through them a couple weeks ago but didn’t have time - still on the to do list. 😬

LiliC:
@tashimi No worries, I created an issue for it and will get to it asap. Thanks again!

krnowak:
I suppose that in our case there's no difference between none and standalone topology since we do not use any update strategy.

Anyway, this PR LFAD. (looks fine and dandy)

PR #117, opened 2017-10-23
examples: Change webserver port to 30001 in README

asymmetric:
5555 is the port on the Pod, whereas 30001 is the one accessible on the
Node.


asymmetric:
OK, but let's set aside some time to discuss benefit/cost ratio of this approach at some point.

zeenix:
@asymmetric I thought we already did and i recall providing reasons but sure.

PR #118, opened 2017-10-30
*: Add example deployment manifest for operator

LiliC:
To ensure that the appropriate number of replicas of the Habitat
operator are always running, this introduces a Deployment rather than a
single static Pod.

### DIFF
--- a/examples/habitat-operator-deployment.yml
+++ b/examples/habitat-operator-deployment.yml
@@ -1,0 +1,14 @@
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: habitat-operator
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        name: habitat-operator
+    spec:
+      containers:
+      - name: habitat-operator
+        image: kinvolk/habitat-operator:v0.1.0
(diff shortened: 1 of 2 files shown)


examples/habitat-operator-deployment.yml
```diff
@@ -0,0 +1,14 @@
...
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        name: habitat-operator
+    spec:
+      containers:
+      - name: habitat-operator
+        image: kinvolk/habitat-operator:v0.1.0
```
krnowak: Maybe latest? But if we decide to keep it, then we need to remember to bump this version everytime we do a release.

LiliC: I would actually tag it with specific version, rather then `latest`. Yes that should be part of the release process. Will maybe add some docs about the release process soon, wdyt?

krnowak: Yeah. Could you file an issue about it? Please mention the need for updating the version number on each release.

LiliC: Issue filed https://github.com/kinvolk/habitat-operator/issues/119, feel free to add anything you can think of there.

PR #125, opened 2017-11-08
Fix Habitat removal

### LINKED ISSUE #122: Errors when deleting Habitat resource
When deleting a Habitat, the following errors are displayed:

```
ts=2017-11-08T12:02:17.98051371+01:00 level=info component=controller msg="deleted deployment" name=example-leader-follower-habitat
ts=2017-11-08T12:02:17.983844937+01:00 level=error component=controller msg="deployments.apps \"example-leader-follower-habitat\" not found"
ts=2017-11-08T12:02:17.98428139+01:00 level=error component=controller msg="Habitat could not be synced, requeueing" msg="deployments.apps \"example-leader-follower-habitat\"
 not found"
```

The actual Habitat is removed.

This seems like it could have been introduced in #113.

asymmetric:
This fixes #122.

We were enqueueing Deployments as well as Habitats, but then expecting Habitats in the `enqueue` function.

Additionally, we now perform a type assertion on every handler, for consistency and safety.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -261,11 +261,18 @@ func (hc *HabitatController) handleHabDelete(obj interface{}) {
 func (hc *HabitatController) handleDeployAdd(obj interface{}) {
 	d, ok := obj.(*appsv1beta1.Deployment)
 	if !ok {
-		level.Error(hc.logger).Log("msg", "Failed to type assert deployment", "obj", obj)
+		level.Error(hc.logger).Log("msg", "Failed to type assert Deployment", "obj", obj)
 		return
 	}
+
 	if isHabitatObject(&d.ObjectMeta) {
-		hc.enqueue(obj)
+		h, err := hc.getHabitatFromResource(d)
+		if err != nil {
+			level.Error(hc.logger).Log("msg", "Could not find Habitat for Deployment", "name", d.Name)
+			return
+		}
+
+		hc.enqueue(h)
 	}
 }
 
@@ -277,7 +284,13 @@ func (hc *HabitatController) handleDeployUpdate(oldObj, newObj interface{}) {
 	}
 
 	if isHabitatObject(&d.ObjectMeta) {
-		hc.enqueue(newObj)
+		h, err := hc.getHabitatFromResource(d)
+		if err != nil {
+			level.Error(hc.logger).Log("msg", "Could not find Habitat for Deployment", "name", d.Name)
+			return
+		}
+
+		hc.enqueue(h)
 	}
 }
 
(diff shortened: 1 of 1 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -865,10 +897,9 @@ func newConfigMap(ip string) *apiv1.ConfigMap {
 	}
 }
 
-func habitatKeyFromPod(pod *apiv1.Pod) string {
-	hName := pod.Labels[crv1.HabitatNameLabel]
-
-	key := fmt.Sprintf("%s/%s", pod.Namespace, hName)
+func habitatKeyFromResource(r metav1.Object) string {
```
LiliC: We can't get key from all resources, as they are not all labeled, so I feel like the name is a bit misleading.

pkg/habitat/controller/controller.go
```diff
@@ -865,10 +897,9 @@ func newConfigMap(ip string) *apiv1.ConfigMap {
...
+	hName := r.GetLabels()[crv1.HabitatNameLabel]
```
LiliC: Should we check to make sure the label we get is not an empty string? WDYT?

pkg/habitat/controller/controller.go
```diff
@@ -706,10 +738,10 @@ func (hc *HabitatController) newDeployment(h *crv1.Habitat) (*appsv1beta1.Deploy
 	return base, nil
 }
 
-func (hc *HabitatController) enqueue(obj interface{}) {
-	k, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+func (hc *HabitatController) enqueue(hab *crv1.Habitat) {
+	k, err := cache.DeletionHandlingMetaNamespaceKeyFunc(hab)
 	if err != nil {
-		level.Error(hc.logger).Log("msg", "Object key could not be retrieved", "object", obj)
+		level.Error(hc.logger).Log("msg", "Object key could not be retrieved", "object", hab)
```
LiliC: I guess since we changed it to `Habitat`. We can do something like:
`"msg", "Habitat key could not be retrieved"`

pkg/habitat/controller/controller.go
asymmetric: I'll changed it to `..FromLabelResource`

asymmetric: Or [labeled](http://grammarist.com/spelling/label/) :confused:

pkg/habitat/controller/controller.go
LiliC: SGTM.

pkg/habitat/controller/controller.go
LiliC: haha :D

My two cents: I think American English tends to win in these cases.

pkg/habitat/controller/controller.go
```diff
@@ -855,6 +890,19 @@ func (hc *HabitatController) getHabitatFromPod(pod *apiv1.Pod) (*crv1.Habitat, e
 	return h, nil
 }
 
+// habitatKeyFromLabeledResource returns a Store key for any resource tagged
+// with the `HabitatNameLabel`.
+func habitatKeyFromLabeledResource(r metav1.Object) (string, error) {
+	hName := r.GetLabels()[crv1.HabitatNameLabel]
```
LiliC: We should add an extra check here for `exists`, to prevent panic when a map would access a non existing key.

asymmetric: There is no panic when the key doesn't exist. The zero-value is [returned instead](https://golang.org/doc/effective_go.html#maps):

> An attempt to fetch a map value with a key that is not present in the map will return the zero value for the type of the entries in the map.

PR #128, opened 2017-11-10
Add channels support to the Habitat type

### LINKED ISSUE #126: Add channels support to the Habitat type
Habitat applications can exist in the different "channels" - like "staging", "production", "unstable" etc. Those names can be configured by the user.

Those channels will be expressed as labels in Kubernetes.

This idea requires adding the additional field in Habibat type - `Channel string`

nhlfr:
Habitat applications can exist in the different "channels" -
like "staging", "production", "unstable" etc. Those names
can be configured by the user.

Those channels will be expressed as labels in Kubernetes.

There is also a possibiity of promoting Habitat applications.
This action is expressed by HabitatPromote object, which needs
to be created in order to perform it.

To test that functionality, you can use the example yml files:

    kubectl create -f examples/standalone/habitat.yml
    kubectl create -f examples/standalone/habitat-promote.yml

Fixes #126

Signed-off-by: Michal Rostecki <michal@kinvolk.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,12 @@ To create an example service run:
 This will create a single-pod deployment of an `nginx` Habitat service.
 More examples are located in the [example directory](https://github.com/kinvolk/habitat-operator/tree/master/examples/).
 
+### Promoting a Habitat application
+
+By default, the example service from `examples/standalone/habitat.yml` uses the `staging` channel. To promote it to the `production` channel, use the following example:
+
+    kubectl create -f examples/standalone/habitat-promote.yml
+
 ## Contributing
 
 ### Dependency management
--- a/docs/api.md
+++ b/docs/api.md
@@ -35,3 +35,19 @@ The following is a description of the Habitat operator API. To see manifest exam
 | name | Name of the bind specified in the Habitat configuration files. | string | true |
 | service | Name of the service this bind refers to. | string | true |
 | group | Group of the service this bind refers to. | string | true |
+
+## HabitatPromote
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| metadata |  | [metav1.ObjectMeta](https://kubernetes.io/docs/api-reference/v1.6/#objectmeta-v1-meta) | true |
+| spec |  | [HabitatSpec](#habitatspec) | true |
+| status |  |  | false |
+
+## HabitatPromoteSpec
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| habitatName | Name of the Habitat object to promote. | string | true |
+| oldChannel | Name of the channel the Habitat application is on before the promotion. | string | true |
+| newChannel | Name of the channel the Habitat application is moved to as part of the promotion. | string | true |
--- a/pkg/habitat/apis/cr/v1/types.go
+++ b/pkg/habitat/apis/cr/v1/types.go
@@ -29,6 +29,12 @@ const (
 	HabitatNameLabel = "habitat-name"
 
 	TopologyLabel = "topology"
+
+	// HabitatChannelLabel contains the information about stability of application.
+	// Example: 'channel: production'
+	HabitatChannelLabel = "channel"
+
+	HabitatPromotionResourcePlural = "habitatpromotions"
 )
 
 type Habitat struct {
@@ -44,6 +50,9 @@ type HabitatSpec struct {
 	// Image is the Docker image of the Habitat Service.
 	Image   string  `json:"image"`
 	Service Service `json:"service"`
+	// Channel is the information about stability of the application, expressed as a label in Kubernetes.
+	// Optional.
+	Channel string `json:"channel,omitempty"`
 }
 
 type HabitatStatus struct {
@@ -94,8 +103,34 @@ const (
 	TopologyLeader     Topology = "leader"
 )
 
+type HabitatPromotion struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+	Spec              HabitatPromotionSpec   `json:"spec"`
+	Status            HabitatPromotionStatus `json:"status,omitempty"`
+}
+
+type HabitatPromotionSpec struct {
+	HabitatName string `json:"habitatName"`
+	OldChannel  string `json:"oldChannel"`
+	NewChannel  string `json:"newChannel"`
+}
+
+type HabitatPromotionStatus struct {
+	State   HabitatPromotionState `json:"state,omitempty"`
+	Message string                `json:"message,omitempty"`
+}
+
+type HabitatPromotionState string
+
 type HabitatList struct {
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata"`
 	Items           []Habitat `json:"items"`
+}
+
+type HabitatPromotionList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata"`
+	Items           []HabitatPromotion `json:"items"`
 }
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -56,6 +56,9 @@ const (
 	// Keys are saved to disk with the format `<name>-<revision>.<extension>`.
 	// This regexp captures the name part.
 	ringKeyRegexp = `^([\w_-]+)-\d{14}$`
+
+	// The default channel for habitat applications, if not specified.
+	defaultHabChannel = "unstable"
 )
 
 var ringRegexp *regexp.Regexp = regexp.MustCompile(ringKeyRegexp)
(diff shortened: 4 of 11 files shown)


pkg/habitat/apis/cr/v1/types.go
```diff
@@ -29,6 +29,10 @@ const (
 	HabitatNameLabel = "habitat-name"
 
 	TopologyLabel = "topology"
+
+	// HabitatChannelLabel contains the information about stability of application.
+	// Examole: 'channel: production'
```
dongsupark: NIT: s/Examole/Example/

pkg/habitat/controller/controller.go
```diff
@@ -578,9 +578,10 @@ func (hc *HabitatController) newDeployment(h *crv1.Habitat) (*appsv1beta1.Deploy
 			Template: apiv1.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
 					Labels: map[string]string{
-						crv1.HabitatLabel:     "true",
-						crv1.HabitatNameLabel: h.Name,
-						crv1.TopologyLabel:    topology.String(),
+						crv1.HabitatLabel:        "true",
```
LiliC: I am a bit confused why choose labels, maybe I am missing some info? From what I can tell you can pass in the argument the same way we do for other options: `--channel unstable`?

nhlfr: I don't understand the question. I'm passing the label to the deployment just to store the info (from which channel comes the app).

LiliC: What I meant, will these labels be only used for information purposes?

pkg/habitat/apis/cr/v1/types.go
```diff
@@ -44,6 +48,8 @@ type HabitatSpec struct {
 	// Image is the Docker image of the Habitat Service.
 	Image   string  `json:"image"`
 	Service Service `json:"service"`
+	// Channel is the information about stability of the application, expressed as a label in Kubernetes.
+	Channel string `json:"channel"`
```
asymmetric: Is this a mandatory field? If not, please add optional to the comment.

Would it make sense to have a default value when the user doesn't provide one?

robertgzr: Habitats default seems to be `unstable`

pkg/habitat/controller/controller.go
nhlfr: Well, the other purpose may be creating services for specific environments (to expose ports). When you have an application with one name, but with two channels (i.e. staging and production), you can create two services with selectors differing in channel.

tashimi:
All newly built packages are automatically put on `unstable` to indicate there is a new build. A user then must do a promotion command (e.g. `hab pkg promote foo`) to promote a package to a different channel. 

You can attach a label to a group of services, Application.Environment, with the idea being you can use this as an organizational principle to do group promotion of various tiers of your application. This is a capability we haven't fully fleshed out/made available yet. 

For now, you should have the ability to promote a package from `unstable` to `foo` to `stable`. All packages by default will have at least an `unstable` and `stable` channel. Additional channels are user defined and added, and there can be as many as you want to power the promotion experience you want.

nhlfr:
PTALA

LiliC:
@nhlfr it seems your tests failed, I re-trigged the build. Will have a look once they pass. :)

docs/api.md
```diff
@@ -35,3 +35,19 @@ The following is a description of the Habitat operator API. To see manifest exam
...
+| metadata |  | [metav1.ObjectMeta](https://kubernetes.io/docs/api-reference/v1.6/#objectmeta-v1-meta) | true |
+| spec |  | [HabitatSpec](#habitatspec) | true |
+| status |  |  | false |
+
+## HabitatPromoteSpec
+
+| Field | Description | Scheme | Required |
+| ----- | ----------- | ------ | -------- |
+| habitatName | Name of the Habitat object to promote. | string | true |
+| oldChannel | Name of the old channel which the Habitat application is using. | string | true |
```
asymmetric: I suggest: "Name of the channel the Habitat application is on before the promotion."

docs/api.md
```diff
@@ -35,3 +35,19 @@ The following is a description of the Habitat operator API. To see manifest exam
...
+| newChannel | Name of the new channel which the Habitat application is going to use after promotion. | string | true |
```
asymmetric: I suggest: "Name of the channel the Habitat application is moved to as part of the promotion."

examples/standalone/habitat-promote.yml
```diff
@@ -0,0 +1,8 @@
+apiVersion: habitat.sh/v1
+kind: HabitatPromote
```
asymmetric: Shouldn't it be `HabitatPromotion`? Isn't that more correct for a type/kind, i.e. that it is a noun?

README.md
```diff
@@ -61,6 +61,12 @@ To create an example service run:
 This will create a single-pod deployment of an `nginx` Habitat service.
 More examples are located in the [example directory](https://github.com/kinvolk/habitat-operator/tree/master/examples/).
 
+### Promoting an example
+
+By default, the example service from `examples/standalone/habitat.yml` uses the `staging` channel. To promote it to the `production` channel, use the following example:
+
+    kubectl create -f examples/standalone/habitat-promote.yml
```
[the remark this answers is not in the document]
nhlfr: But that specific file is able to promote only the **standalone** application, it cannot be used to any other example!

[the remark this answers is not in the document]
asymmetric: The `examples/promotion` directory would contain both the habitat app's manifest and the promotion manifest to promote that app. Or am I missing something?

nhlfr:
@asymmetric ptala

pkg/habitat/controller/controller.go
```diff
@@ -243,8 +275,19 @@ func (hc *HabitatController) handleHabUpdate(oldObj, newObj interface{}) {
 		return
 	}
 
-	if hc.habitatNeedsUpdate(oldHab, newHab) {
-		hc.enqueue(newHab)
+	if hc.habitatIsPromotiond(oldHab, newHab) {
```
asymmetric: Is this a typo?

pkg/habitat/controller/controller.go
```diff
@@ -243,8 +275,19 @@ func (hc *HabitatController) handleHabUpdate(oldObj, newObj interface{}) {
...
+		// Remove old deployment
```
asymmetric: You forgot the very important `.` at the end! ;)

pkg/habitat/controller/controller.go
```diff
@@ -243,8 +275,19 @@ func (hc *HabitatController) handleHabUpdate(oldObj, newObj interface{}) {
...
+		deploymentName := hc.deploymentName(oldHab.Name, oldHab.Spec.Channel)
+		if err := hc.config.KubernetesClientset.AppsV1beta1Client.Deployments(apiv1.NamespaceDefault).Delete(deploymentName, &metav1.DeleteOptions{}); err != nil {
+			level.Error(hc.logger).Log("msg", "Failed to delete deployment", "name", deploymentName)
+			return
+		}
+		level.Info(hc.logger).Log("msg", "created deployment", "name", deploymentName)
```
asymmetric: Is this the correct message? Should it mention promotion, as in "promoted deployment"?

pkg/habitat/controller/controller.go
```diff
@@ -845,6 +953,15 @@ func (hc *HabitatController) conform(key string) error {
 	return nil
 }
 
+func (hc *HabitatController) habitatIsPromotiond(oldHabitat, newHabitat *crv1.Habitat) bool {
```
asymmetric: Typo, should be "HabitatIsPromoted"

pkg/habitat/controller/controller.go
```diff
@@ -748,45 +856,45 @@ func (hc *HabitatController) enqueue(hab *crv1.Habitat) {
...
 	}
 
-	hc.queue.Add(k)
+	hc.habitatQueue.Add(k)
 }
 
-func (hc *HabitatController) worker() {
-	for hc.processNextItem() {
+func (hc *HabitatController) habitatWorker() {
+	for hc.processNextHabitatItem() {
```
asymmetric: All these renames make me think we should split the promotion into its own controller. If this PR is urgent and blocking we can postpone the split, though.

asymmetric:
Also, tests are failing.

nhlfr:
Needs to be re-designed and re-implemented

nhlfr:
I'm not deleting the branch just to keep the code visible

asymmetric:
What's the discussion that led to this decision?

robertgzr:
we talked about it offline but the basic outcome was that a promotion should just be another
`kubectl apply -f manifest.yml`

For that the channel probably has to be part of the object name:
hab pkg ident: `origin/pkg/ver/release`, in channel `unstable`
becomes
`origin-pkg-ver-release-unstable` object (no need for random strings)

We don't have to remove the old deployment since that can be easily scripted if labels are set properly...

Issue #119, opened 2017-10-30
Release process


LiliC:
There should be a documented release process, i.e. steps we need to take for when doing a new release of the Habitat operator. Here are a few steps that come to my mind right now:

- Build image `make linux`
- Tag image to the release following versioning strategy `vx.x.x` 
- Push the image to the version tag as well as `latest` to [hub.docker.com](https://hub.docker.com/r/kinvolk/habitat-operator/)
- Bump up tag version in the deployment Habitat operator manifest file `examples/habitat-operator-deployment.yml`
- Update `CHANGELOG.md` with release notes

PR #130, opened 2017-11-21
Document release process

LiliC:
This PR adds a bit of automation to our release process, it adds `VERSION` file where the latest version of the Habitat operator is written. It also adds `update-version` to the Makefile which makes use of the `VERSION` file and replaces the tagged images in the examples.

Also added some documentation for the release process, these are the steps I did in the last two releases.

Closes https://github.com/kinvolk/habitat-operator/issues/119.


Makefile
```diff
@@ -25,4 +25,8 @@ e2e:
 clean-test:
 	kubectl delete namespace testing
 
-.PHONY: build test linux image e2e clean-test
+update-version:
+	sed -i.bak 's/habitat-operator:.*/habitat-operator:v$(shell cat VERSION)/g' examples/habitat-operator-deployment.yml
```
asymmetric: Won't we be left with those `.bak` files after running this?

Also, I'm not sure we need backups at all, since it's all under VCS anyway, so I suggest just going with `-i` for in-place editing.

release.md
```diff
@@ -0,0 +1,38 @@
+# Release process
+
+Habitat operator follows [semantic versioning](https://semver.org/) e.g. `1.2.3`.
```
asymmetric: I don't think we need the `e.g. 1.2.3`, also because IMO it doesn't really explain what SemVer is.

release.md
```diff
@@ -0,0 +1,38 @@
...
+## Tag the release
+
+Tag the new release with a tag named v<major>.<minor>.<patch>, e.g. `v2.1.3`, and push the tag.
+
+    git tag -a vx.y.z -m 'vx.y.z'
+    git push vx.y.z
+
+## Generate release image
+
+In the root directory of the repository generate the Docker image and push it to the Kinvolk Docker hub:
```
asymmetric: I think "to the Kinvolk repo on Docker Hub" is more correct.

release.md
```diff
@@ -0,0 +1,38 @@
...
+First bump the version in the `VERSION` file which is located in the top level directory of the repository.
+Run the following command to regenerate certain files:
+
+    make update
+
+Document the changes in this release in the `CHANGELOG.md` file, following the already established pattern. Commit all the changes, including the regenerated files and `CHANGELOG.md` file under one commit message e.g.: `*: cut 0.2.0 release`. Create a PR with the changes.
+
+## Create release branch
+
+After the above mentioned changes were merged into master, locally create a branch with a new major or minor release if one doesn't already exist. Otherwise cherry-pick the changes to the already existing minor release branch.
```
asymmetric: > After the above mentioned changes were merged

"have been merged" sounds more correct IMO.

asymmetric: > locally create a branch with a new major or minor release if one doesn't already exist. Otherwise cherry-pick the changes to the already existing minor release branch.

I'm not entirely clear what the process is when the branch already exists. When does that happen? Shouldn't every release produce a new release number?

Can we add more info about this, or is there some link we can point people (and myself 😄) to?

release.md
```diff
@@ -0,0 +1,38 @@
...
+
+    git checkout -b release-<major>.<minor>
```
asymmetric: So the branch name won't include the patch version? Is this also an established workflow coming from SemVer?

Makefile
krnowak: - I'd really prefer something that does not hardcode the list of files where the version needs to be replaced. I'd prefer something like getting a list of files from `git ls-files` and act on them. Likely with excluding the vendor directory.
  - The reason is that otherwise we would need to remember to copy the sed invocation for every added file that contains the version number of the habitat-operator. And the list of those files may grow as we add more manifest examples and documentation.
- Also, would be nice to replace the version without requiring `habitat-operator:` to be in the same line - that may not be true for some documentation we may add in future.
- The `-i.bak` is not necessary - you are going to have a nice diff with `git diff`. So just `-i` is enough.
- Consider replacing `$(shell cat VERSION)` with `$$(cat version)` and using double quotes instead of single quotes.

release.md
```diff
@@ -0,0 +1,38 @@
+# Release process
+
+Habitat operator follows [semantic versioning](https://semver.org/) e.g. `1.2.3`.
+
+## Pull request
+
+First bump the version in the `VERSION` file which is located in the top level directory of the repository.
+Run the following command to regenerate certain files:
+
+    make update
```
krnowak: It's `make update-version`.

LiliC:
@krnowak 
> make commit-and-upload-release, it would do:
git commit …
git tag …
git push …
build docker image and push it.

Well you first need to open the PR and have it merged before you can proceed with the rest of the release process, so doing `make commit-and-upload-release` can't be done. And automating just the `git tag` part is a bit pointless IMHO.

> I think that doing release branches is pointless - we don't do fixes to branches for now.

I think it's a good habit to have, and it doesn't hurt anyone?

release.md
LiliC: Sure will do, it's what upstream Kubernetes does and all the other operators so I followed that pattern.

krnowak:
@LiliC 
> > make commit-and-upload-release, it would do:
> > git commit …
> > git tag …
> > git push …
> > build docker image and push it.

> Well you first need to open the PR and have it merged before you can proceed with the rest of the release process, so doing make commit-and-upload-release can't be done. And automating just the git tag part is a bit pointless IMHO.

Then work flow would be after manually checking if version updating didn't do anything wrong:

- commit changes
- file a PR
- if it gets merged, you could have `make tag-and-upload`, which would do basically the same things:
  - git tag
  - git push (to push the tags to repo)
  - build docker image and push it

So it would not be automating only the git tag parts. `make tag-and-upload` would take the version from the already updated VERSION file.

> > I think that doing release branches is pointless - we don't do fixes to branches for now.
> 
> I think it's a good habit to have, and it doesn't hurt anyone.

It's pointless, because:
- We are likely not going to make patch level releases any time soon. And release branches are basically a set of all the patch versions of a given major-minor release.
  - We could very well stick to the major-minor versioning, without the patch level, but I'm not going to insist on that.
- We didn't promise any stability yet. So in this case even the patch level release could be breaking. That would make release branches pointless.

LiliC:
> We are likely not going to make patch level releases any time soon. 

@krnowak actually, we would need this now. ;) As with https://github.com/kinvolk/habitat-operator/pull/133 merged, we can patch v0.2 branch. WDYT?

krnowak:
I still think that using release branches is pointless unless we are going to develop on several release branches in parallel.

Instead of doing the patch release, consider making another regular one.

LiliC:
@krnowak I am for having it, as its a common practice in both upstream Kubernetes as well as the other operators. But will let @asymmetric make the final vote. 😃

release.md
asymmetric: I also think it would be good to explicitly give a schema for the branch name, e.g.:
> create a branch named `release-x.y`

asymmetric:
I think not having release branches simplifies the workflow somehow.
We can use tags as pointers for releases, and always merge to master, including bug-fixes.

If a bugfix warrants a new-release, we create a new tag. i.e. I don't understand why the decision of having release branches should affect our decision of making patch releases at all. 

That sounds simplest to me. Am I missing something?

release.md
zeenix: Still noob at this but I i can understand why a very big project in k8s would want a branch for each and every release (even though I believe the GNOME's "a branch for each release cycle" is the way to go for that) but for this small a project, I don't see the need. Just need a git tag.

zeenix: Also tags should be IMO on master branch commits so it's easy to follow the timeline.

LiliC:
SGTM on the not having the branches.

@asymmetric @zeenix One thing, I was thinking should we add the `release.md` to the `docs` directory or leave it in the root directory?

zeenix:
root dir seems more close to general convention to me.

Makefile
```diff
@@ -25,4 +25,8 @@ e2e:
 clean-test:
 	kubectl delete namespace testing
 
-.PHONY: build test linux image e2e clean-test
+update-version:
+	sed -i '' "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" 'examples/habitat-operator-deployment.yml'
```
asymmetric: This doesn't work on GNU sed. But `sed -i` doesn't work on FreeBSD sed.

Since sed is [apparently not the right tool](http://mywiki.wooledge.org/BashFAQ/021#Using_nonstandard_tools) for this anyway, and this option is not portable, I suggest we use `ex`:

`ex -sc 's/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g|x' examples/habitat-operator-deployment.yml`.

Also, you don't need the single quotes around the filename.

Makefile
```diff
@@ -25,4 +25,8 @@ e2e:
...
+	sed -i '' "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" 'examples/rbac/habitat.yml'
```
asymmetric: @krnowak suggested we don't hard-code the paths, and instead use something like `git ls-files`, and I agree that that seems more maintainable and less brittle.

release.md
```diff
@@ -0,0 +1,32 @@
...
+
+    make update-version
+
+Document the changes in this release in the `CHANGELOG.md` file, following the already established pattern. Commit all the changes, including the regenerated files and `CHANGELOG.md` file under one commit message e.g.: `*: cut 0.2.0 release`. Create a PR with the changes.
+
+## Tag the release
+
+Tag the new release with a tag named v<major>.<minor>.<patch>, e.g. `v2.1.3`, and push the tag.
+
+    git tag -a vx.y.z -m 'vx.y.z'
```
asymmetric: +1 for annotated tags.

Makefile
LiliC: That did not work for me, sadly. It just loops endlessly. 🤔

asymmetric: Yep, same for me, not sure why. I think it has something to do with the fact that `ex` on my machine is actually `nvim`.

How do we deal with this? On that page it also mentions `ed`, but I haven't tried it.

LiliC: What worked fine for me was:
 `sed -i.bak "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" examples/habitat-operator-deployment.yml` 
But that would mean needing to remove the backups in the next step.

zeenix: As said IRL, the sed commands work fine on my Linux machine, as expected, without the empty input to `-i` and .back.

asymmetric: @zeenix I don't understand that. Can you paste the command that works for you?

release.md
```diff
@@ -0,0 +1,32 @@
...
+First bump the version in the `VERSION` file which is located in the top level directory of the repository.
+Run the following command to regenerate certain files:
+
+    make update-version
+
+Document the changes in this release in the `CHANGELOG.md` file, following the already established pattern. Commit all the changes, including the regenerated files and `CHANGELOG.md` file under one commit message e.g.: `*: cut 0.2.0 release`. Create a PR with the changes.
+
+## Tag the release
+
+Tag the new release with a tag named v<major>.<minor>.<patch>, e.g. `v2.1.3`, and push the tag.
```
asymmetric: We should mention that tagging should happen on master, after the PR has been merged.

release.md
```diff
@@ -0,0 +1,32 @@
...
+Habitat operator follows [semantic versioning](https://semver.org/).
+
+## Pull request
+
+First bump the version in the `VERSION` file which is located in the top level directory of the repository.
+Run the following command to regenerate certain files:
+
+    make update-version
+
+Document the changes in this release in the `CHANGELOG.md` file, following the already established pattern. Commit all the changes, including the regenerated files and `CHANGELOG.md` file under one commit message e.g.: `*: cut 0.2.0 release`. Create a PR with the changes.
```
asymmetric: Actually, by the time the PR is created, the [Full changelog link](https://github.com/kinvolk/habitat-operator/compare/v0.2.0...v0.3.0) in the CHANGELOG won't work, because there won't yet be any `v0.x.0` tag, unless the person who makes the release creates a branch with than name.

But I wouldn't recommend that, as then it would be hard to distinguish it from the tag.

Makefile
LiliC: Done. 👍

release.md
LiliC: I am aware of that. But I don't think it's a problem, there is no way around it from what I can see. And like you mentioned its only that way until it's merged.

release.md
LiliC: Done.

Makefile
zeenix: `sed -i "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" 'examples/habitat-operator-deployment.yml'`

release.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Tag the release
+
+After the above mentioned PR was merged, switch to updated master branch. Tag the new release with a tag named v<major>.<minor>.<patch>, e.g. `v2.1.3`, and push the tag.
+
+    git tag -a vx.y.z -m 'vx.y.z'
+    git push vx.y.z
```
asymmetric: This should be `git push origin vx.y.z`, otherwise it won't work.

LiliC: Done.

VERSION
```diff
@@ -0,0 +1 @@
+0.2.0
```
asymmetric: This needs to be updated to `0.3.0` :)

release.md
```diff
@@ -0,0 +1,32 @@
+# Release process
+
+Habitat operator follows [semantic versioning](https://semver.org/).
```
asymmetric: Nit: Usually capitalized to Semantic Versioning.

release.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Pull request
+
+First bump the version in the `VERSION` file which is located in the top level directory of the repository.
```
asymmetric: Nit: "version file, located.."

release.md
```diff
@@ -0,0 +1,32 @@
...
+Run the following command to regenerate certain files:
```
asymmetric: Nit: No files are being regenerated, I'd say "to update the version across the repo"

release.md
asymmetric: Nit: I'd remove regenerated, and just say "Commit all changes and the `CHANGELOG.md` file..."

release.md
```diff
@@ -0,0 +1,32 @@
...
+
+    make update-version
+
+Document the changes in this release in the `CHANGELOG.md` file, following the already established pattern. Commit all the changes, including the regenerated files and `CHANGELOG.md` file under one commit message e.g.: `*: cut 0.2.0 release`. Create a PR with the changes.
+
+## Tag the release
+
+After the above mentioned PR was merged, switch to updated master branch. Tag the new release with a tag named v<major>.<minor>.<patch>, e.g. `v2.1.3`, and push the tag.
```
asymmetric: "switch to the". Also, why "updated"?

release.md
```diff
@@ -0,0 +1,32 @@
...
+In the root directory of the repository generate the Docker image and push it to the Kinvolk repo on Docker hub:
+
+    make image
+    docker push kinvolk/habitat-operator:vx.y.z
+    docker tag kinvolk/habitat-operator:vx.y.z kinvolk/habitat-operator:latest
+    docker push kinvolk/habitat-operator:latest
+
+## Do the release
+
+Head over to github and edit the release notes with the notes that were included in the `CHANGELOG.md` file. Also include the generated Docker image in the release notes.
```
asymmetric: Nit: capitalized GitHub.

release.md
LiliC: Well you just merged the PR, so hopefully you will do a `git pull`...

release.md
LiliC: Including name of the file better describes it, leaving it like this.

LiliC:
@asymmetric  Done.

release.md
asymmetric: Oh yeah, I was only suggesting removing "which is" and adding a comma.

release.md
asymmetric: Ah I see :)

Issue #131, opened 2017-11-22
Crash on startup


zeenix:
I'm seeing a crash on startup:

```
#  ./habitat-operator --kubeconfig /var/lib/kube-spawn/default/kubeconfig 
ts=2017-11-22T16:05:17.159334625+01:00 level=info msg="Habitat CRD already exists, continuing"
ts=2017-11-22T16:05:17.160388968+01:00 level=info component=controller msg="Watching Habitat objects"
ERROR: logging before flag.Parse: E1122 16:05:17.171033    3611 runtime.go:66] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:72
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:65
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:51
/usr/lib/golang/src/runtime/asm_amd64.s:514
/usr/lib/golang/src/runtime/panic.go:489
/usr/lib/golang/src/runtime/panic.go:63
/usr/lib/golang/src/runtime/signal_unix.go:290
<autogenerated>:18
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/vendor/k8s.io/client-go/tools/cache/store.go:84
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/vendor/k8s.io/client-go/tools/cache/controller.go:258
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/pkg/habitat/controller/controller.go:710
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/pkg/habitat/controller/controller.go:330
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/pkg/habitat/controller/controller.go:215
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/vendor/k8s.io/client-go/tools/cache/controller.go:192
<autogenerated>:54
/home/zeenix/golang/src/github.com/kinvolk/habitat-operator/vendor/k8s.io/client-go/tools/cache/shared_informer.go:547
/usr/lib/golang/src/runtime/asm_amd64.s:2197
[...]

PR #133, opened 2017-11-27
Prevent panic when passing nil object

LiliC:
Sometimes it occurred that hab object was nil, and that triggered a
panic. We now check for nil and return instead.

Fixes https://github.com/kinvolk/habitat-operator/issues/131.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -739,6 +739,11 @@ func (hc *HabitatController) newDeployment(h *crv1.Habitat) (*appsv1beta1.Deploy
 }
 
 func (hc *HabitatController) enqueue(hab *crv1.Habitat) {
+	if hab == nil {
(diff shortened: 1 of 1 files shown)


zeenix:
@asymmetric sounds right but I think it'd be nice to have this guard anyway.

LiliC:
I agree with @zeenix on this one, I don't think it hurts having it in that place, unless I am missing something?

dongsupark:
What I meant in https://github.com/kinvolk/habitat-operator/issues/131 was actually that `hc.enqueue()` should be called outside the clause `if err != nil { }`. Then we could avoid the panic gracefully. Obviously we would then need to change error handling again, etc.

So I think this PR is not bad, because it's simple and handling nil case inside `enqueue` does not hurt.

LiliC:
@asymmetric @zeenix @dongsupark I adjusted the PR to include the returning on all errors, instead of enqueuing. I also left the check for `nil`. PTALA, thanks!

PR #135, opened 2017-12-13
Stop workers when controller stops

asymmetric:
This makes sure that all workers are terminated.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -105,6 +105,9 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
 
 // Run starts a Habitat resource controller.
 func (hc *HabitatController) Run(ctx context.Context) error {
+	// Make sure the work queue is shutdown which will trigger workers to end.
+	defer hc.queue.ShutDown()
+
 	level.Info(hc.logger).Log("msg", "Watching Habitat objects")
 
 	hc.cacheHab()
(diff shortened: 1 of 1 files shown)


zeenix:
why not? :) LGTM.

pkg/habitat/controller/controller.go
```diff
@@ -757,10 +760,13 @@ func (hc *HabitatController) worker() {
 }
 
 func (hc *HabitatController) processNextItem() bool {
+	// Process an item, unless `queue.ShutDown()` has been called, in which case we exit.
 	key, quit := hc.queue.Get()
 	if quit {
 		return false
 	}
+
+	// Indicate to the queue that we've completed a unit of work.
```
LiliC: Not sure we need these comments, as `queue.Done(key)` is very descriptive.

pkg/habitat/controller/controller.go
```diff
@@ -778,6 +784,7 @@ func (hc *HabitatController) processNextItem() bool {
 		return true
 	}
 
+	// If there was no error, tell the queue it can stop tracking failure history for the key.
```
LiliC: Same here as above, not sure this comment is needed.

pkg/habitat/controller/controller.go
zeenix: Now that you mention it, yeah I also think these comments come close to what Linus calls over-documentation. :)

pkg/habitat/controller/controller.go
asymmetric: I disagree about this one. It's not clear what `Forget` does exactly without having to dig through k8s' code, and the [`Forget`](https://github.com/kinvolk/habitat-operator/tree/3daa1e86d3e581c107044982de11507e9c896c3b/vendor/k8s.io/client-go/util/workqueue/default_rate_limiters.go#L117) method itself isn't documented.

PR #136, opened 2017-12-13
Start multiple workers

### LINKED ISSUE #67: Increase number of workers
We can add a parameter to control how many workers are started. These workers will pop jobs from the workqueue in parallel, improving performance.

See [this](https://github.com/kubernetes/client-go/blob/9675d048204baf818bb24f7c74d0e64f6254902e/examples/workqueue/main.go#L131-L133) for an example implementation.

asymmetric:
With this PR, we start `runtime.NumCPU()` workers instead of just one. As mentioned in the commit, we might want to, in the future, make this value configurable.

Closes #67.

### DIFF
--- a/cmd/habitat-operator/main.go
+++ b/cmd/habitat-operator/main.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"os"
 	"os/signal"
+	"runtime"
 	"syscall"
 
 	"github.com/go-kit/kit/log"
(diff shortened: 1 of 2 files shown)


asymmetric:
I would argue that when you only need to add a trivial amount of code to get parallelism, it doesn't really need to be justified that much.

The benefit is parallelism in the job processing. What's the downside with merging this change exactly?

zeenix:
fwiw, I agree with @asymmetric 's last comment.

PR #138, opened 2017-12-14
Change log level about IP removal to debug

LiliC:
The "removed peer IP from ConfigMap" msg was not useful in log level
info. Let me know what you think? I found it non informative and a bit spammy.

### DIFF
--- a/pkg/habitat/controller/controller.go
+++ b/pkg/habitat/controller/controller.go
@@ -480,7 +480,7 @@ func (hc *HabitatController) handleConfigMap(h *crv1.Habitat) error {
 				return err
 			}
 
-			level.Info(hc.logger).Log("msg", "removed peer IP from ConfigMap", "name", newCM.Name)
+			level.Debug(hc.logger).Log("msg", "removed peer IP from ConfigMap", "name", newCM.Name)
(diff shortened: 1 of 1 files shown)


pkg/habitat/controller/controller.go
```diff
@@ -480,7 +480,7 @@ func (hc *HabitatController) handleConfigMap(h *crv1.Habitat) error {
 				return err
 			}
 
-			level.Info(hc.logger).Log("msg", "removed peer IP from ConfigMap", "name", newCM.Name)
+			level.Debug(hc.logger).Log("msg", "removed peer IP from ConfigMap", "name", newCM.Name)
```
zeenix: Just a random nit: Why don't we capitalize here? Cause we know that there will be a prefix attached?

asymmetric: Basically because we are outputting a structured log, rather than one immediately meant for user consumption.

Something that ingests these logs higher up in the chain could parse extract and capitalize.

I guess that's the idea anyway. In practice, these logs are mostly viewed by humans at the moment.

I'm also not 100% sure we are consistent with this.

PR #141, opened 2017-12-19
Switch to little-endian dates in CHANGELOG

asymmetric:
Little endian dates are [much more common][0] across the globe, whereas
big-endian ones are almost only used in the USA.

[0]: https://en.wikipedia.org/wiki/Date_format_by_country


asymmetric:
The important things first! ;)

PR #142, opened 2017-12-19
Avoid installing dependencies when building image

asymmetric:
When building an image, we want to avoid the side effect of installing binaries.

Moreover, the `-i` flag required `sudo` on my machine (not sure if it's a general problem though).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ build:
 
 linux:
 	# Compile statically linked binary for linux.
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build --ldflags="-s" -o habitat-operator github.com/kinvolk/habitat-operator/cmd/habitat-operator
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -ldflags="-s" -o habitat-operator github.com/kinvolk/habitat-operator/cmd/habitat-operator
 
 image: linux
 	docker build -t "$(IMAGE):$(TAG)" .
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ build:
 
 linux:
 	# Compile statically linked binary for linux.
-	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -i --ldflags="-s" -o habitat-operator github.com/kinvolk/habitat-operator/cmd/habitat-operator
+	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build --ldflags="-s" -o habitat-operator github.com/kinvolk/habitat-operator/cmd/habitat-operator
 
 image: linux
 	docker build -t "$(IMAGE):$(TAG)" .


asymmetric:
From `go help build`:

> The -i flag installs the packages that are dependencies of the target.

LiliC:
@asymmetric https://ttboj.wordpress.com/2017/02/26/faster-golang-builds/

asymmetric:
I guess that's a side-effect of the files being there. Still, running this on a default install of go on Ubuntu requires `sudo`, which means that the instructions are wrong.

I think not requiring `sudo` is more important than the time saving. WDYT?

asymmetric:
The point is that this make target shouldn't have side effects imho, because it's not meant to install anything on the host, it's meant to build a binary for the image.

Proof is that it builds for Linux even on OSX. So I'd say that the installation is an unwanted side-effect, that trumps the **nice** side-effect of speeding up compilation.

But the whole point is IMO moot because we should build the image in the container  using [multi-stage builds](https://medium.com/travis-on-docker/multi-stage-docker-builds-for-creating-tiny-go-images-e0e1867efe5a) ;).

LiliC:
> But the whole point is IMO moot because we should build the image in the container using multi-stage builds ;).

Think I would prefer that, rather then changing this here. Can you maybe open a ticket for it?

LiliC:
Sorry I didn't mean to close :/

krnowak:
The situation with GOPATH and GOROOT stinks, and it makes `-i` flag crippled. Basically, most of the time, the artifacts built with `go build -i` will land somewhere below `${GOPATH}/pkg`, so in a location that is writable. But this is different for the golang standard library as it is not a thing you find in GOPATH.

If GOROOT env var is defined, they will end up in `${GOROOT}/pkg/…`.
If there is no GOROOT env var defined then it probably depends on where did you get your go from:
- from tarball - it will try to put the artifacts in `/usr/local/go/pkg/…`
- from distro - wherever the distro finds it suitable, usually read-only location like `/usr/lib/golang/pkg/…`
- build from git - some magic sauce detects it and puts them in `<GIT_REPO>/pkg/…`.

So for me it worked, since I'm using go built from git. What we could do in this case is (as we don't have that multi stage docker whatnot):
```bash
gorootdir=$(go env | grep '^GOROOT=' | sed -e 's/.*"\([^"]*\)"/\1/'); \
if test -w "${gorootdir}"; then i_flag='-i'; fi; \
GOOS=… GOARCH=… go build ${i_flag} …
```

asymmetric:
`make linux` without the `-i` flag takes ~22seconds on my machine. Even if the `-i` flag makes that faster, that's not enough of a time saving for me to justify @krnowak's workaround. I'd just remove the `-i` flag.

Interestingly, `make build` doesn't require sudo on my machine. It must be one of the additional flags we pass to `go build` that triggers it.

krnowak:
A bit more findings.

`make build` not requiring `sudo` is because distros usually provide prebuilt artifacts for most common cases. In my case I have 3 sets of them, one for normal builds (`linux_amd64`, that's what we do in `make build`), one for normal builds with race detector enabled (`linux_amd64_race`) and one that would imply something with dynamic linking (`linux_amd64_dynlink`, no clue what's its for). So during `make build` you don't build the standard library and you don't install the artifacts in GOROOT, but instead reuse the prebuilt objects distro provided to you.

In case of `make linux` we pass a bunch of special stuff (I guess its both `CGO_ENABLED=0` and `-ldflags="-s"`), so `go build -i` figures out that the prebuilt artifacts are incompatible, so it rebuilds the standard library, but fails at installing the artifacts in GOROOT because permissions.

As a side note: we would need to provide `-installsuffix=<some suffix>` flag along with `-i` in `make linux` so we don't try to replace artifacts for the normal builds with ones built with special flags. That of course won't fix the issue of permissions, though. The `-installsuffix` is not needed for `make build` as it is doing the normal build, so it should reuse the prebuilt artifacts anyway.

I also did a bit of benchmarking. Results for doing `make linux` with `-i` flag removed three times:
```
real	0m20.299s
user	0m58.713s
sys	0m3.499s

real	0m19.758s
user	0m58.635s
sys	0m3.466s

real	0m19.950s
user	0m58.783s
sys	0m3.424s
```

So basically around 20 seconds with low deviation.

Results for `make linux` with `-i -installsuffix=nocgo_static` run 4 times:
```
real	0m20.415s
user	0m59.180s
sys	0m3.692s

real	0m1.581s
user	0m1.803s
sys	0m0.299s

real	0m1.519s
user	0m1.675s
sys	0m0.314s

real	0m1.530s
user	0m1.716s
sys	0m0.305s
```

So subsequent rebuilds go down to mere 1,5 second.

Results for the `make linux` with making changes in `pkg/controller/controller.go`:
```
real	0m2.010s
user	0m2.412s
sys	0m0.313s

real	0m2.029s
user	0m2.417s
sys	0m0.320s

real	0m1.991s
user	0m2.366s
sys	0m0.335s
```

So in general, it's quite a nice improvement (from 20 seconds down to 2 or 1,5).

So, I'd keep `make build` as it is now, but we can drop the `-i` flag from `make linux` because it is only used for building the docker images. And we will lose the benefit of `-i` anyway if we end up doing this multi-stage building with docker.

And about side-effects: this point is moot IMO also because of the way go was designed (GOPATH and GOROOT shenanigans, meh).

Issue #145, opened 2017-12-22
Conform to upstream api structure


LiliC:
Currently our structure does not match the [upstream api directory structure](https://github.com/kubernetes/sample-controller/tree/release-1.9/pkg). This breaks when trying to automate updates between versions. For example when running the generated script introduced in v1.8.0, it fails to generate the correct files as we place our apis and client dirs t in different subdirectories.

The benefits of switching are:
- we will conform to the same structure as other upstream controllers as well as other operators
- we can automate version updates and not need to write the generated code manually

PR #147, opened 2018-01-02
Conform to upstream dir layout structure

LiliC:
Currently our directory structure is not aligned with upstream Kubernetes API directory layout. The upstream code generator tools make assumptions on the specific directory layout.

I also renamed the `cr` to `habitat`, as this makes more sense now that the `habitat` directory is gone. And it is also what all the upstream and other operators do.

Closes https://github.com/kinvolk/habitat-operator/issues/145.


pkg/client/client.go
```diff
@@ -16,7 +16,8 @@
 package client
 
 import (
-	crv1 "github.com/kinvolk/habitat-operator/pkg/habitat/apis/cr/v1"
+	habv1 "github.com/kinvolk/habitat-operator/pkg/apis/habitat/v1"
```
LiliC: One thing I would like to get feedback on:
I think we should move to `v1beta1`, as `v1` is considered stable and I am not sure we are there yet with our API. The change doesn't need to happen in this PR. Just collecting feedback here.

schu: Nit: in other files you use `habitatv1`, not `habv1`. Maybe change to `habitatv1` here.

LiliC: Actually we already had naming pattern established with `habitat/client` -> `habclient`, so I just changed this to `habv1` instead.

(Sorry for the force push.)

krnowak: The v1beta looks good, I think. But that would need some planning when we can declare the API as stable.

cmd/habitat-operator/main.go
```diff
@@ -29,8 +29,8 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 
-	habitatclient "github.com/kinvolk/habitat-operator/pkg/habitat/client"
-	habitatcontroller "github.com/kinvolk/habitat-operator/pkg/habitat/controller"
+	habitatclient "github.com/kinvolk/habitat-operator/pkg/client"
+	habitatcontroller "github.com/kinvolk/habitat-operator/pkg/controller"
```
krnowak: habclient and habcontroller?

LiliC:
@krnowak Addressed your comment. PTALA.

Issue #108, opened 2017-10-06
e2e tests: use yaml files instead of API calls


LiliC:
To catch errors such as the one in https://github.com/kinvolk/habitat-operator/pull/104, we should switch in the e2e tests to yaml manifest files, instead of creating our operator SG resources through the API. As we would have noticed the problem of the tag not being changed properly.

Issue #148, opened 2018-01-02
Run e2e tests against RBAC enabled Kubernetes cluster


LiliC:
Currently we run our end-to-end tests against a non RBAC enabled cluster.

asymmetric:
@LiliC was this closed by #151?

Issue #105, opened 2017-10-05
Decrease e2e tests running time


asymmetric:
End-to-end tests currently run for 10+ minutes.

Ideas:

* Disable Travis' PR test
* Find a way to only run certain tests sometime

LiliC:
Closing as the above PR was merged.

Issue #121, opened 2017-11-08
Remove dependency on minikube


asymmetric:
The `e2e` target in the Makefile [calls](https://github.com/kinvolk/habitat-operator/blob/371bbbad65de6aa8826c04fcd144deef833fa4a3/Makefile#L23) `minikube ip`.

It would be good not to expect a minikube installation to be present, and get the cluster IP some other way, e.g. by parsing `~/.kube/config`.

zeenix:
Doesn't `--kubeconfig ~/.kube/config` part just before --ip option in the Makefile takes care of that?

asymmetric:
It doesn't take care of that, but it contains the information we need, so one option is to parse that file instead of calling `minikube ip`.

krnowak:
Probably this could be used:
- `kubectl config view --output=jsonpath='{.clusters[0].cluster.server}' | sed -e 's/.*\/\/\([0-9.]\+\).*/\1/'`
- `kubectl config view --output=jsonpath='{.clusters[0].cluster.server}' | grep --only-matching '[0-9.]\+' | head --lines 1`

Second one is easier to understand I think.

asymmetric:
See #266: we can still use `minikube` locally but E2E tests will be run on Azure in the future during our CI process.

So this issue is still valid.

PR #151, opened 2018-01-11
Switch to e2e testing against RBAC enabled cluster and refactor tests

LiliC:
This PR introduces testing against a RBAC enabled cluster. I also introduce taking the manifest files from the examples rather then creating the resources through the API in code. See issue https://github.com/kinvolk/habitat-operator/issues/108 for more information.

Also bumped the version of cluster to 1.8.0.

Closes https://github.com/kinvolk/habitat-operator/issues/108 and https://github.com/kinvolk/habitat-operator/issues/105 and https://github.com/kinvolk/habitat-operator/issues/148.

This improved the e2e tests from aprox. 257.289s seconds down to 121.449 seconds. 🎉

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,10 +17,10 @@ git:
 
 before_script:
 # Download kubectl, which is a requirement for using minikube.
-- curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.7.0/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+- curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.8.0/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
 # Download minikube.
 - curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
-- sudo minikube start --vm-driver=none --kubernetes-version=v1.7.0
+- sudo minikube start --vm-driver=none --kubernetes-version=v1.8.0 --extra-config=apiserver.Authorization.Mode=RBAC
 # Fix the kubectl context, as it's often stale.
 - minikube update-context
 # Wait for Kubernetes to be up and ready.
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -17,7 +17,6 @@
 package framework
 
 import (
-	habv1 "github.com/kinvolk/habitat-operator/pkg/apis/habitat/v1"
 	habclient "github.com/kinvolk/habitat-operator/pkg/client"
 
 	apiv1 "k8s.io/api/core/v1"
@@ -101,8 +100,7 @@ func (f *Framework) setupOperator() error {
 		return err
 	}
 
-	// Wait until the operator is ready.
-	f.WaitForResources(name, 1)
+	f.WaitForResources("name", d.ObjectMeta.Name, 1)
 
 	return nil
 }
--- a/test/e2e/framework/framework.go
+++ b/test/e2e/framework/framework.go
@@ -80,26 +80,23 @@ func Setup(image, kubeconfig, externalIP string) (*Framework, error) {
 }
 
 func (f *Framework) setupOperator() error {
-	name := "habitat-operator"
-	pod := &apiv1.Pod{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: name,
-			Labels: map[string]string{
-				habv1.HabitatNameLabel: name,
-			},
-		},
-		Spec: apiv1.PodSpec{
-			Containers: []apiv1.Container{
-				{
-					Name:  name,
-					Image: f.Image,
-				},
-			},
-		},
+	// Setup RBAC for operator.
+	err := f.createRBAC()
+	if err != nil {
+		return err
 	}
 
-	// Create pod with the Habitat operator image.
-	_, err := f.KubeClient.CoreV1().Pods(TestNs).Create(pod)
+	// Get Habitat operator deployment from examples.
+	d, err := ConvertDeployment("resources/operator/habitat.yml")
+	if err != nil {
+		return err
+	}
+
+	// Override image with the one passed to the tests.
+	d.Spec.Template.Spec.Containers[0].Image = f.Image
+
+	// Create deployment for the Habitat operator.
+	_, err = f.KubeClient.AppsV1beta1().Deployments(TestNs).Create(d)
 	if err != nil {
 		return err
 	}
--- a/test/e2e/framework/helpers.go
+++ b/test/e2e/framework/helpers.go
@@ -25,37 +25,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
-// NewStandaloneHabitat returns a new Standalone Habitat.
-func NewStandaloneHabitat(habitatName, group, image string) *habv1.Habitat {
-	return &habv1.Habitat{
-		ObjectMeta: metav1.ObjectMeta{
-			Name: habitatName,
-		},
-		Spec: habv1.HabitatSpec{
-			Image: image,
-			Count: 1,
-			Service: habv1.Service{
-				Group:    group,
-				Topology: habv1.TopologyStandalone,
-			},
-		},
-	}
-}
-
-// AddConfigToHabitat adds a ConfigSecretName field to the Habitat.
-func AddConfigToHabitat(habitat *habv1.Habitat) {
-	habitat.Spec.Service.ConfigSecretName = habitat.ObjectMeta.Name
-}
-
-// AddBindToHabitat appends bind fields to the Habitat.
-func AddBindToHabitat(habitat *habv1.Habitat, bindName, bindService string) {
-	habitat.Spec.Service.Bind = append(habitat.Spec.Service.Bind, habv1.Bind{
-		Name:    bindName,
-		Service: bindService,
-		Group:   habitat.Spec.Service.Group,
-	})
-}
-
 // CreateHabitat creates a Habitat.
 func (f *Framework) CreateHabitat(habitat *habv1.Habitat) error {
 	return f.Client.Post().
--- a/test/e2e/framework/helpers.go
+++ b/test/e2e/framework/helpers.go
@@ -15,14 +15,21 @@
 package framework
 
 import (
+	"os"
+	"path/filepath"
 	"time"
 
 	habv1 "github.com/kinvolk/habitat-operator/pkg/apis/habitat/v1"
 
+	appsv1beta1 "k8s.io/api/apps/v1beta1"
+	"k8s.io/api/core/v1"
+	apiv1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/util/yaml"
 )
 
 // CreateHabitat creates a Habitat.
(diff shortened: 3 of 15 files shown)


LiliC:
@schu Thanks for having a look! Agree we should make it independat of the tool you use to setup your k8s cluster...

> to be not limited to the K8s versions offered by minikube when testing..

Fair point, but at least locally for testing I use `--bootstrapper kubeadm` to test latest versions of K8s and that seems to work well with minikube. There is not need for the operator to work with master of K8s or anything like that.

As for switching from minikube I personally didn't have much need for it right now, as I tend to often just push my branch and let travis do the rest for me, so no need to even have minikube locally. What I would love to see is a multi node K8s setup for testing, maybe not on every PR, but I would love to have that automated at some point, which is out of the scope of minikube. But the options are sort of limited right now.... 🙂 But maybe you can open an issue or contribute your thoughts to the already open issue https://github.com/kinvolk/habitat-operator/issues/121, that would be greatly appreciated! 👍

schu:
> Fair point, but at least locally for testing I use --bootstrapper kubeadm to test latest versions of K8s and that seems to work well with minikube

Cool, didn't know that's possible.

test/e2e/framework/helpers.go
```diff
@@ -117,6 +93,178 @@ func (f *Framework) DeleteHabitat(habitatName string) error {
...
+// It converts it from that file to the ServiceAccount object.
+func ConvertServiceAccount(pathToYaml string) (*apiv1.ServiceAccount, error) {
+	manifest, err := PathToOSFile(pathToYaml)
+	if err != nil {
+		return nil, err
+	}
+
+	sa := apiv1.ServiceAccount{}
+
+	if err := yaml.NewYAMLOrJSONDecoder(manifest, 100).Decode(&sa); err != nil {
```
krnowak: The function docs (and the parameter name implies that too) that the passed file is yaml, so you could use `NewYAMLToJSONDecoder` here instead, right? That would also avoid passing this `100` number which prompted me to search what is the meaning of this number.

krnowak: Same in all those dull repeated functions, because no generics for us. ;)

test/e2e/framework/helpers.go
```diff
@@ -117,6 +93,178 @@ func (f *Framework) DeleteHabitat(habitatName string) error {
...
+		return err
+	}
+
+	return nil
+}
+
+// ConvertServiceAccount takes in a path to the YAML file containing the manifest
+// It converts it from that file to the ServiceAccount object.
+func ConvertServiceAccount(pathToYaml string) (*apiv1.ServiceAccount, error) {
+	manifest, err := PathToOSFile(pathToYaml)
```
krnowak: So, these functions are indeed repetitive. One way to handle that would probably be to have a function like:

```go
func convertToK8sFoo(pathToYaml string, into interface{}) -> error {
    // call pathToOSFile (or inline it here)
    // create yaml decoder and return a result of the call to Decode(into)
}
```

then these functions could be like:

```go
func ConvertServiceAccount(pathToYaml string) (*apiv1.ServiceAccount, error) {
    sa := apiv1.ServiceAccount{}
    if err := convertToK8sFoo(pathToYaml, &sa); err != nil {
        return nil, err
    }
    return &sa, nil
}
```

test/e2e/framework/helpers.go
```diff
@@ -117,6 +93,178 @@ func (f *Framework) DeleteHabitat(habitatName string) error {
...
+	if err := yaml.NewYAMLOrJSONDecoder(manifest, 100).Decode(&s); err != nil {
+		return nil, err
+	}
+
+	return &s, nil
+}
+
+// PathToOSFile takes in a path and converts it to a File.
+func PathToOSFile(relativPath string) (*os.File, error) {
+	path, err := filepath.Abs(relativPath)
```
krnowak: I think getting an absolute path here is pointless, isn't it?

Also, `s/relativPath/relativePath/`.

test/e2e/framework/framework.go
```diff
@@ -80,26 +80,23 @@ func Setup(image, kubeconfig, externalIP string) (*Framework, error) {
...
+	// Setup RBAC for operator.
+	err := f.createRBAC()
+	if err != nil {
+		return err
 	}
 
-	// Create pod with the Habitat operator image.
-	_, err := f.KubeClient.CoreV1().Pods(TestNs).Create(pod)
+	// Get Habitat operator deployment from examples.
+	d, err := ConvertDeployment("../../examples/rbac/habitat.yml")
```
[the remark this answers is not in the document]
krnowak: Eh, scratch those comments, I wrote more about it in the main comment.

.travis.yml
```diff
@@ -17,10 +17,10 @@ git:
 
 before_script:
 # Download kubectl, which is a requirement for using minikube.
-- curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.7.0/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+- curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.8.0/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
```
asymmetric: Not sure this should be part of this PR, or is there a reason?

examples/bind-config/habitat-go.yml
```diff
@@ -0,0 +1,13 @@
+apiVersion: habitat.sh/v1
```
asymmetric: What's the advantage of splitting these two? I think it made more sense before, since IIUC they are meant to be tried out together.

examples/rbac/README.md
```diff
@@ -4,9 +4,12 @@
 
 ## Workflow
 
-Before deploying the Habitat operator inside your cluster the following roles must be created:
+Before deploying the Habitat operator inside your cluster the following resources must be created:
+
+    kubectl apply -f examples/rbac/service-account.yml
```
asymmetric: Again, I see a preference for splitting up that I don't necessarily share, but that's OK :)

But maybe, If that's all the files in the dir, it would be more user-friendly to just indicate so and use `kubectl apply -f examples/rbac`, cause otherwise it gets tedious really fast.

test/e2e/framework/helpers.go
```diff
@@ -117,6 +93,178 @@ func (f *Framework) DeleteHabitat(habitatName string) error {
...
+	if err := yaml.NewYAMLOrJSONDecoder(manifest, 100).Decode(&sa); err != nil {
+		return nil, err
+	}
+
+	return &sa, nil
+}
+
+// ConvertClusterRole takes in a path to the YAML file containing the manifest.
+// It converts the file to the ClusterRole object.
+func ConvertClusterRole(pathToYaml string) (*rbacv1.ClusterRole, error) {
```
asymmetric: This function (and a few others) doesn't seem to be used outside, so maybe it should be private.

test/e2e/operator_test.go
```diff
@@ -36,167 +35,69 @@ const (
...
-	}
-
-	// Wait for resources to be ready.
-	if err := framework.WaitForResources(habitatName, 1); err != nil {
-		t.Fatal(err)
-	}
-
-	_, err := framework.KubeClient.CoreV1().ConfigMaps(utils.TestNs).Get(configMapName, metav1.GetOptions{})
+// TestFunction tests that the operator correctly created two Habitat Services and bound them together.
+func TestFunction(t *testing.T) {
```
asymmetric: Can't we use a more descriptive name? TestBinding?

examples/bind-config/habitat-go.yml
asymmetric: Ah, I see now that it's because this one is used from inside the tests. Maybe we can just say what this is/does.

test/e2e/operator_test.go
```diff
@@ -212,46 +113,51 @@ func TestHabitatFunctioning(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// This msg is set in the default.toml in kinvolk/nodejs-hab Habitat Service.
-	expectedMsg := "Hello, World!"
+	// This msg is set in the config of the kinvolk/bindgo-hab Go Habitat Service.
+	expectedMsg := "hello from port: 4444"
```
asymmetric: If we change the port in the example and forget to change this, we'll have a failing test, which means we still have to keep the two in sync, which is kind of what we were trying to avoid, isn't it?

Maybe we shoulnd't match against the port part.

.travis.yml
LiliC: @asymmetric Did it here as I was already touching that part of the "code". Think we can slowly move away from v1.7.x considering v1.9.x is out. 

Note: The codebase was tested with 1.7.x and 1.9.x.

examples/rbac/README.md
LiliC: > Again, I see a preference for splitting up that I don't necessarily share, but that's OK :)

WDYT of then creating these files which will be used in tests in the `tests/e2e/resources/...` directory?

test/e2e/operator_test.go
LiliC: The only problem is that we are doing this to test that the binding worked. But agreed, so if you agree with my suggestion above to move these under the tests directory?

test/e2e/framework/framework.go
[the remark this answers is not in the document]
krnowak: Sounds good.

LiliC:
@krnowak @asymmetric Think I either responded or addressed all your comments, can you have another look, thanks!

test/e2e/operator_test.go
asymmetric: Yes, I agree with loading the files under `examples/` in our tests.

examples/rbac/README.md
asymmetric: Ah, this is a new idea! So would these be just copies of the files under `examples/`? If yes, I don't see the advantage.

LiliC: See @krnowak comment above:
> How about leaving the examples as they used to be and having a separate set of manifests with rbac and bind config stuff only for testing? That would introduce a bit of redundancy, true, but it would keep the examples simple and without hidden strings like "it is being used for our testing".

asymmetric: I'm 👎 about that. I think the concern about tying examples and tests too abstract, i.e. I fail to see what the issue could be.

Also, would these generated example fixtures just be copies of the test files? Or would they be generated in code? If it's the former, I don't see the point; If it's the latter, that would be even more complexity and make even less sense IMO. 🤷‍♂️

test/e2e/framework/helpers.go
```diff
@@ -117,6 +93,156 @@ func (f *Framework) DeleteHabitat(habitatName string) error {
...
+// convertClusterRoleBinding takes in a path to the YAML file containing the manifest.
+// It converts the file to the ClusterRoleBinding object.
+func convertClusterRoleBinding(pathToYaml string) (*rbacv1.ClusterRoleBinding, error) {
+	crb := rbacv1.ClusterRoleBinding{}
+
+	if err := convertToK8sResource(pathToYaml, &crb); err != nil {
+		return nil, err
+	}
+
+	// Change namespace to testing.
```
krnowak: Is this still necessary, given that the tests now have their own set of manifests?

test/e2e/resources/operator/cluster-role-binding.yml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: habitat-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: habitat-operator
+subjects:
+- kind: ServiceAccount
+  name: habitat-operator
+  namespace: default
```
krnowak: This namespace can be different now, no?

examples/rbac/README.md
krnowak: What I wanted to have is to leave examples as is - simple single files as manifests. And I wanted them to have a single purpose - just to serve as examples. So the tests should have their own set of manifests and they can be split however you like them to be.

Of course, it would be nice to have some examples validator (so it's a valid yaml or maybe even valid manifest), but that's it - no checking if operator behaves correctly.

LiliC:
@krnowak good catch, fixed. Can you have another look please!

test/e2e/operator_test.go
krnowak: The port changes sounds like a good argument for having a separate set of manifests for testing, so innocuous changes in some example file won't break the tests. So that's what we have settled for now in this PR - not touching the examples at all, and having a bunch of manifests for testing.

krnowak:
LFAD if green.

examples/rbac/README.md
LiliC: Agree with @krnowak on this one, I initially used the examples manifest files as well. But I think having them under the tests is better for the above reasons.

We can add another set of tests for the examples, but that would be done outside of this PR.

PR #155, opened 2018-01-17
Add Name field

asymmetric:
This PR adds the `Habitat.Spec.Service.Name` mandatory field.

The field is used by the operator to know in which directory under `/hab/svc/` to place the `user.toml` file.

This also allows users to use any name they want for the `ObjectMeta.Name` field, e.g. `myapp-v1`, making it possible to run multiple versions of the same Service.


examples/namespaced/habitat.yml
```diff
@@ -6,13 +6,14 @@ metadata:
...
 metadata:
-  name: example-standalone-habitat
+  name: example-namespaced-habitat
   namespace: example-namespace
 spec:
   # the core/nginx habitat service packaged as a Docker image
   image: kinvolk/nginx-hab
   count: 1
   service:
+    name: nginx
```
LiliC: Considering the field is optional I wouldn't use it in the examples that don't have a secret? WDYT? If you agree it applies for all the other examples.

pkg/apis/habitat/v1/types.go
```diff
@@ -61,16 +61,19 @@ type Service struct {
...
+	// Optional.
+	UserConfigSecretName string `json:"userConfigSecretName,omitempty"`
 	// The name of the secret that contains the ring key.
 	// Optional.
 	RingSecretName string `json:"ringSecretName,omitempty"`
 	// Bind is when one service connects to another forming a producer/consumer relationship.
 	// Optional.
 	Bind []Bind `json:"bind,omitempty"`
+	// Name is the name of the Habitat service that this Habitat object represents.
+	// This field is used, among other things, to know where in the /hab/svc/ hierarchy to place a user.toml file.
```
LiliC: "Among other things..." What are those?

LiliC: I would rephrase it. Currently this field is used to place the `user.toml` in the correct directory. WDYT?

examples/namespaced/habitat.yml
asymmetric: It's not optional at the moment, but open for discussion. I think making it optional makes it more error prone.

Right now we only need it if there's additional user config, but in the future it could be used for other things, and then the user would have to remember what those are.

Also, someone could start with a manifest without config, then add config but forget to add the name..

It just seems better to make it mandatory IMO.

asymmetric:
About the tests PR, I had forgotten it was in the pipeline. I think I'll wait until it's merged.

pkg/apis/habitat/v1/types.go
asymmetric: "This field is used to place the user.toml file in the correct subdirectory of /hab/svc/ in the Pod."?

pkg/controller/controller.go
```diff
@@ -713,7 +713,7 @@ func (hc *HabitatController) newDeployment(h *habv1.Habitat) (*appsv1beta1.Deplo
 
 		secretVolumeMount := &apiv1.VolumeMount{
 			Name: initialConfigFilename,
-			// Our user.toml file must be in a directory with the same name as the service.
+			// The Habitat supervisor creates a directory for each service under /hab/svc/<servicename>. We need to place the user.toml file in there in order for it to be detected.
```
zeenix: is there no max chars limit in Go world?

asymmetric: [Nope](https://golang.org/doc/effective_go.html#formatting):

> Go has no line length limit. Don't worry about overflowing a punched card. If a line feels too long, wrap it and indent with an extra tab.

But I agree the line was too long, fixed.

asymmetric:
PTALA.

pkg/apis/habitat/v1/types.go
LiliC: Yes that sgtm.

pkg/controller/controller.go
```diff
@@ -695,25 +697,26 @@ func (hc *HabitatController) newDeployment(h *habv1.Habitat) (*appsv1beta1.Deplo
...
 			},
 		}
 
 		secretVolumeMount := &apiv1.VolumeMount{
-			Name: "initialconfig",
-			// Our user.toml file must be in a directory with the same name as the service.
-			MountPath: fmt.Sprintf("/hab/svc/%s/%s", h.Name, userTomlFile),
-			SubPath:   userTomlFile,
+			Name: initialConfigFilename,
+			// The Habitat supervisor creates a directory for each service under /hab/svc/<servicename>.
```
LiliC: Nitpick: 

Think this is a bit clearer:
> The Habitat supervisor creates a directory for each service under /hab/svc, like so /hab/svc/<servicename>.
> The user.toml file needs to be placed in that directory in order for Habitat to pick up the configuration and applies them.

PR #157, opened 2018-01-23
Use cache for ConfigMaps

### LINKED ISSUE #156: Cache for ConfigMaps
Use a cache instead of making `GET` calls to the API.

asymmetric:
This PR adds a cache for ConfigMaps, and uses that - instead of calling the API directly - whenever we want to perform a `GET`. Writes and updates obviously still go to the API.

It also does a couple other things:

* Sets up all watchers to only returned labelled objects
* Removes manual checks for the label in the handlers, because see above

I decided not to implement a cache for the secrets because I'm not convinced it makes sense in this case: in the other situations, we check the cache only when creating an object fails. By that time, we have created an object, so we have the necessary `struct`.

In the case of Secrets, we assume they have been created by the user beforehand, so we shouldn't have to create one just so that we have something to pass to `MetaNamespaceKeyFunc`.

Closes #156.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -172,11 +172,11 @@ func (hc *HabitatController) cacheHabitats() {
 }
 
 func (hc *HabitatController) cacheDeployments() {
-	source := cache.NewListWatchFromClient(
+	source := newListWatchFromClientWithLabels(
 		hc.config.KubernetesClientset.AppsV1beta1().RESTClient(),
 		"deployments",
 		apiv1.NamespaceAll,
-		fields.Everything())
+		labelListOptions())
 
 	hc.deployInformer = cache.NewSharedIndexInformer(
 		source,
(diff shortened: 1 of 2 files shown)


asymmetric:
PTALA.

asymmetric:
PTALAA

pkg/controller/controller.go
```diff
@@ -868,10 +868,6 @@ func (hc *HabitatController) conform(key string) error {
 		} else {
 			return err
 		}
-		_, err = hc.config.KubernetesClientset.AppsV1beta1().Deployments(h.Namespace).Get(deployment.Name, metav1.GetOptions{})
```
krnowak: The result being unused/ignored is not true here, contrary to what the commit message says ("The output of the call was being ignored."). We still check the error.

asymmetric: That's true. But I wonder, what's the point of `GET`ing the Deployment? We just checked that it already exists.

krnowak: Yes, please. That would at least explain the reason behind the change.

asymmetric: Done.

asymmetric:
PTALA.

pkg/controller/controller.go
krnowak: Hm, I don't see anything changed…

asymmetric:
PTALA.

PR #159, opened 2018-01-24
[WIP] Downgrade Trusty image

### LINKED ISSUE #158: Tests fail on latest Travis image
The failure is about [not being able to connect](https://travis-ci.org/kinvolk/habitat-operator/jobs/332745907#L598) to the Service that should be listening on `30001`.

The only workaround is to downgrade to the previous version of the trusty image, as [explained here](https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch).

asymmetric:
This is a workaround for the issue where we can't connect to the Service during tests.

Closes #158.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: go
 
+dist: trusty
 sudo: required
 
 # This moves Kubernetes specific config files.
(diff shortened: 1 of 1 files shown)


LiliC:
We can close this now, as it seems like they solved their problems at Travis, but will keep the branch just in case. ;)

asymmetric:
Yes, I think we should still extract the versions into a variable for discoverability.

PR #160, opened 2018-01-25
vendor: clean up unused imports

LiliC:
With the new release of dep, it is easy to purge on each dep ensure.
This commit also adds the necessary configs to the Gopkg.lock file to be
able to do that.


krnowak:
Is there anything that checks that dep is 0.4 or newer? I wonder how dep will behave if somebody uses an older version.

Otherwise LFAD if green.

LiliC:
@krnowak good point, will add it to the README. That we require dep version onwards from `0.4`.

LiliC:
@krnowak done. 👍

asymmetric:
FYI the [`[solve-meta]`](https://golang.github.io/dep/docs/Gopkg.lock.html) section of the `Gopkg.lock` has versioning information. In our case, it contains [this](https://github.com/kinvolk/habitat-operator/blob/dcba88e4663bc67f2f1715344679a35ca7510c22/Gopkg.lock#L427-L432).

PR #161, opened 2018-01-26
Add helm chart for the operator

zeenix:
This is the source for the Helm chart. You can use:

helm package helm/habitat-operator

command to create a chart package for the operator for distribution.

This chart uses the docker image of the latest release: 0.4.0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,13 @@ clean-test:
 	kubectl delete clusterrole habitat-operator
 
 update-version:
-	find examples -name "*.yml" -type f -exec sed -i.bak "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" '{}' \;
-	find examples -name "*.yml.bak" -type f -exec rm '{}' \;
+	find examples helm -name "*.yml" -o -name "*.yaml" -type f \
+		-exec sed -i.bak \
+		-e "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" \
+		-e "s/tag:.*/tag: v$$(cat VERSION)/g" \
+		-e "s/version:.*/version: $$(cat VERSION)/g" \
+		'{}' \;
+	find examples helm -name "*.yml.bak" -o -name "*.yaml.bak" -type f \
+		-exec rm '{}' \;
 
 .PHONY: build test linux image e2e clean-test update-version
--- a/helm/habitat-operator/Chart.yaml
+++ b/helm/habitat-operator/Chart.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+description: A Kubernetes controller that manages Habitat services in a Kubernetes cluster.
+engine: gotpl
+maintainers:
+  - name: Zeeshan Ali
+    email: zeeshan@kinvolk.io
+name: habitat-operator
+sources:
+  - https://github.com/kinvolk/habitat-operator
+version: 0.4.0
--- a/helm/habitat-operator/README.md
+++ b/helm/habitat-operator/README.md
@@ -1,0 +1,109 @@
+# habitat-operator
+
+Installs [habitat-operator](https://github.com/kinvolk/habitat-operator) to manage Habitat services in a Kubernetes cluster.
+
+## TL;DR;
+
+```console
+$ helm repo add habitat https://kinvolk.github.io/habitat-operator/helm/charts/stable/
+$ helm install habitat/habitat-operator
+```
+
+## Introduction
+
+This chart bootstraps a [habitat-operator](https://github.com/kinvolk/habitat-operator) deployment in a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
+
+## Prerequisites
+
+See [Habitat Operator's README](https://github.com/kinvolk/habitat-operator/blob/master/README.md).
+
+### RBAC
+If role-based access control (RBAC) is enabled in your cluster, you may need to give Tiller (the server-side component of Helm) additional permissions. **If RBAC is not enabled, be sure to set `rbacEnable` to `false` when installing the chart.**
+
+1. Create a ServiceAccount for Tiller in the `kube-system` namespace
+```console
+$ kubectl -n kube-system create sa tiller
+```
+
+2. Create a ClusterRoleBinding for Tiller
+
+#### kubectl v1.5.x
+
+```console
+$ cat <<EOF | kubectl create -f -
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+kind: ClusterRoleBinding
+metadata:
+  name: tiller
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+  - kind: ServiceAccount
+    name: tiller
+    namespace: kube-system
+EOF
+```
+
+#### kubectl >= v1.6.x
+
+```console
+$ kubectl create clusterrolebinding tiller --clusterrole cluster-admin --serviceaccount=kube-system:tiller
+```
+
+3. Install Tiller, specifying the new ServiceAccount
+```console
+$ helm init --service-account tiller
+```
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```console
+$ helm install --name my-release habitat/habitat-operator
+```
+
+The command deploys habitat-operator on the Kubernetes cluster in the default configuration. The [configuration](#configuration) section lists the parameters that can be configured during installation.
+
+## Uninstalling the Chart
+
+To uninstall/delete the `my-release` deployment:
+
+```console
+$ helm delete my-release
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following table lists the configurable parameters of the habitat-operator chart and their default values.
+
+Parameter | Description | Default
+--- | --- | ---
+`image.repository` | Image | `docker.io/kinvolk/habitat-operator`
+`image.tag` | Image tag | The latest release tag (e.g `v0.4.0`)
+`image.pullPolicy` | Image pull policy | `IfNotPresent`
+`nodeSelector` | Node labels for pod assignment | `{}`
+`rbacEnable` | If true, create & use RBAC resources | `true`
+`resources` | Pod resource requests & limits | `{}`
+
+Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
+
+```console
+$ helm install --name my-release habitat/habitat-operator --set sendAnalytics=true
+```
+
+Alternatively, a YAML file that specifies the values for the above parameters can be provided while installing the chart. For example,
+
+```console
+$ helm install --name my-release habitat/habitat-operator -f values.yaml
+```
+
+> **Tip**: You can use the default [values.yaml](values.yaml)
+
+## Hosting
+
+The Helm chart repository is managed on the [`gh-pages` branch](https://github.com/kinvolk/habitat-operator/tree/gh-pages) and instructions for repository management can be found [here](https://github.com/kinvolk/habitat-operator/tree/gh-pages/helm/charts).
--- a/helm/habitat-operator/templates/NOTES.txt
+++ b/helm/habitat-operator/templates/NOTES.txt
@@ -1,0 +1,4 @@
+The Habitat Operator has been installed. Check its status by running:
+  kubectl --namespace {{ .Release.Namespace }} get pods -l "app={{ template "habitat-operator.name" . }},release={{ .Release.Name }}"
+
+You can now deploy Habitat services in the cluster.
--- a/helm/habitat-operator/templates/clusterrole.yaml
+++ b/helm/habitat-operator/templates/clusterrole.yaml
@@ -1,0 +1,49 @@
+{{- if .Values.rbacEnable }}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1beta1" }}
+apiVersion: rbac.authorization.k8s.io/v1beta1
+{{- else if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1alpha1" }}
+apiVersion: rbac.authorization.k8s.io/v1alpha1
+{{- else if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1" }}
+apiVersion: rbac.authorization.k8s.io/v1
+{{- end }}
+kind: ClusterRole
+metadata:
+  labels:
+    app: {{ template "habitat-operator.name" . }}
+    chart: {{ .Chart.Name }}-{{ .Chart.Version }}
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+  name: {{ template "habitat-operator.fullname" . }}
+rules:
+- apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups:
+  - habitat.sh
+  resources:
+  - habitats
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups:
+  - apps
+  resources:
+  - deployments
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: [""]
+  resources:
+  - configmaps
+  verbs: ["get", "list", "watch", "create", "update", "patch", "delete"]
+- apiGroups: [""]
+  resources:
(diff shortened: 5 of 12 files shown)


helm/habitat-operator/README.md
```diff
@@ -0,0 +1,100 @@
...
+$ helm install habitat/habitat-operator
+```
+
+## Introduction
+
+This chart bootstraps a [habitat-operator](https://github.com/kinvolk/habitat-operator) deployment on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
+
+## Prerequisites
+  - Habitat `>= 0.36.0`
+  - Kubernetes cluster with version `1.7.x`, `1.8.x` or `1.9.x`.
```
asymmetric: Can't we just say ">- 1.7.x". The above approach doesn't scale ;)

helm/habitat-operator/README.md
```diff
@@ -0,0 +1,100 @@
...
+If role-based access control (RBAC) is enabled in your cluster, you may need to give Tiller (the server-side component of Helm) additional permissions. **If RBAC is not enabled, be sure to set `rbacEnable` to `false` when installing the chart.**
+
+1. Create a ServiceAccount for Tiller in the `kube-system` namespace
+```console
+$ kubectl -n kube-system create sa tiller
+```
+
+2. Create a ClusterRoleBinding for Tiller
+```console
+# kubectl v1.5.x
```
asymmetric: I think you should turn this into a section, e.g. `###`, and have two separate code blocks for each version of k8s.

helm/habitat-operator/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+```console
+$ helm delete my-release
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following tables lists the configurable parameters of the habitat-operator chart and their default values.
```
asymmetric: *table

helm/habitat-operator/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+description: A Kubernetes controller designed to solve running and auto-managing Habitat services on Kubernetes.
+engine: gotpl
+maintainers:
+  - name: Zeeshan Ali
+    email: zeeshan@kinvolk.io
+name: habitat-operator
+sources:
+  - https://github.com/kinvolk/habitat-operator
+version: 0.4.0
```
LiliC: Where did this version come from? Considering this is the first release of this chart this version number should be `0.0.1`?

helm/habitat-operator/Chart.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+description: A Kubernetes controller designed to solve running and auto-managing Habitat services on Kubernetes.
```
LiliC: Nit: Would get rid of the "auto" part here. Would add creating as well.

helm/habitat-operator/README.md
```diff
@@ -0,0 +1,100 @@
+# habitat-operator
+
+Installs [habitat-operator](https://github.com/kinvolk/habitat-operator) to create/configure/manage Habitat services atop Kubernetes.
+
+## TL;DR;
+
+```console
+$ helm repo add habitat https://urltochart.org/
```
LiliC: urltochart.org? :)

helm/habitat-operator/README.md
```diff
@@ -0,0 +1,100 @@
+# habitat-operator
+
+Installs [habitat-operator](https://github.com/kinvolk/habitat-operator) to create/configure/manage Habitat services atop Kubernetes.
```
LiliC: s/atop/inside/g

helm/habitat-operator/README.md
```diff
@@ -0,0 +1,100 @@
...
+## TL;DR;
+
+```console
+$ helm repo add habitat https://urltochart.org/
+$ helm install habitat/habitat-operator
+```
+
+## Introduction
+
+This chart bootstraps a [habitat-operator](https://github.com/kinvolk/habitat-operator) deployment on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
```
LiliC: s/on a/inside of/g

helm/habitat-operator/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Prerequisites
```
LiliC: I wouldn't duplicate these, maybe just link in the operator README section?

helm/habitat-operator/README.md
```diff
@@ -0,0 +1,100 @@
...
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following tables lists the configurable parameters of the habitat-operator chart and their default values.
+
+Parameter | Description | Default
+--- | --- | ---
+`image.repository` | Image | `docker.io/kinvolk/habitat-operator`
+`image.tag` | Image tag | `v0.4.0`
```
LiliC: Can you please add this to the https://github.com/kinvolk/habitat-operator/blob/master/Makefile#L30 So that it is auto updated when doing the release.

helm/habitat-operator/templates/NOTES.txt
```diff
@@ -0,0 +1,4 @@
+The Habitat Operator has been installed. Check its status by running:
+  kubectl --namespace {{ .Release.Namespace }} get pods -l "app={{ template "habitat-operator.name" . }},release={{ .Release.Name }}"
+
+You can now deploy Habitat services through Helm.
```
LiliC: Think you do not need Helm to deploy Habitat services, maybe rephrase this?

helm/habitat-operator/templates/clusterrole.yaml
```diff
@@ -0,0 +1,47 @@
+{{- if .Values.rbacEnable }}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1beta1" }}
```
LiliC: Hmm but now RBAC is not in beta or alpha but it is v1. Maybe add that condition as well?

helm/habitat-operator/templates/clusterrolebinding.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if .Values.rbacEnable }}
+{{- if .Capabilities.APIVersions.Has "rbac.authorization.k8s.io/v1beta1" }}
```
LiliC: Same as above for RBAC.

helm/habitat-operator/templates/deployment.yaml
```diff
@@ -0,0 +1,35 @@
...
+        app: {{ template "habitat-operator.name" . }}
+        operator: habitat
+        release: {{ .Release.Name }}
+    spec:
+      containers:
+        - name: {{ template "habitat-operator.name" . }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+          imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+          ports:
+            - containerPort: 8080
```
LiliC: A bit confused by this? Why is this needed?

helm/habitat-operator/values.yaml
```diff
@@ -0,0 +1,25 @@
+## Habitat-operator image
+##
+image:
+  repository: docker.io/kinvolk/habitat-operator
+  tag: v0.4.0
```
LiliC: Same as above, can you please add this to the https://github.com/kinvolk/habitat-operator/blob/master/Makefile#L30 So that it is auto updated when doing the release.

helm/habitat-operator/README.md
zeenix: sure

helm/habitat-operator/README.md
zeenix: Good idea. Actually this confused me in the prometheus chart and I ended up using the incorrect command.

helm/habitat-operator/Chart.yaml
zeenix: hmm..creating? The operator doesn't create them, does it? To me the "designed to solve" sounds more redundant. How about:

"A Kubernetes controller that manages Habitat services on Kubernetes."

helm/habitat-operator/Chart.yaml
zeenix: Like many other things pointed out in reviews here, a leftover from copy&paste of prometheus operator chart. :)

helm/habitat-operator/README.md
zeenix: since it doesn't say "Kubernetes cluster", "atop" is correct and sounds better to me so I kept that  from the prometheus operator chart.

helm/habitat-operator/README.md
zeenix: exactly! we don't have a URL yet. :) Once we know where to host it, we can set this.

helm/habitat-operator/README.md
zeenix: hmm.. in here it does say "cluster" so that would be more accurate but I don't think "on" is incorrect either and I'd prefer it unless you feel strongly about this.

helm/habitat-operator/README.md
zeenix: you mean link *to* the operator README? sure.

helm/habitat-operator/README.md
zeenix: Sure thing but we really shouldn't be modifying source files. The convention is to have put the source content into .in files and then build can generate the final version.

helm/habitat-operator/templates/NOTES.txt
zeenix: ok

helm/habitat-operator/templates/clusterrole.yaml
zeenix: c&p mistake, I'll just modify it to v1 everywhere.

helm/habitat-operator/templates/deployment.yaml
zeenix: another c&p mistake probably. Will remove or come back with reasons (if any).

helm/habitat-operator/README.md
LiliC: Yes, link to to the operator README.

helm/habitat-operator/README.md
LiliC: Sure, feel free to do that.

helm/habitat-operator/templates/clusterrole.yaml
LiliC: No, I meant this is good, as we are supporting k8s versions of 1.7 and 1.8 as well. But we need to add v1 as well for 1.9+.

helm/habitat-operator/README.md
zeenix: hmm.. I already modified the Makefile rule for existing yaml files but the problem with generating README.md is that then you don't get the nice integration with github. :( Probably best not the mention the actual version here.

zeenix:
PR updated. For the installation, let's for now go with the simplest approach of providing chart tarball as one of the release artifacts. We can look into moving the chart source and tarball to another repo (possibly the main charts repo) later.

zeenix:
Ok, realized that we have to setup a chart repo for being able to add operator as a dep to other charts. I update the PR again and chose a canonical location on kinvok.io where we'll put helm charts for now and will be our repo for now.

I also removed my patch about version update and now version is updated the same as before.

zeenix:
Alright so now even the helm chart repo is setup and I tested the instruction to work. Now would be the best time to PATALA and merge.

asymmetric:
PATALA? 😂

Makefile
```diff
@@ -28,7 +28,13 @@ clean-test:
 	kubectl delete clusterrole habitat-operator
 
 update-version:
-	find examples -name "*.yml" -type f -exec sed -i.bak "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" '{}' \;
-	find examples -name "*.yml.bak" -type f -exec rm '{}' \;
+	find examples helm -name "*.yml" -o -name "*.yaml" -type f \
```
asymmetric: Can't we do `find . -regex '.*\.ya?ml$'`?

Also, the sed invocation was changed quite a bit. Why? I think this could've been its own commit.

helm/habitat-operator/Chart.yaml
asymmetric: So will this be fixed or should it stay like this?

blixtra:
Perhaps it's good to point out how the charts are hosted. Right now the instructions for hosting are only on the gh-pages branch. Wouldn't hurt to at least point to that from the main documentation.
wdyt, @LiliC @asymmetric?

Makefile
zeenix: > Can't we do find . -regex '.*\.ya?ml$'?

That's hard to read so I'll be against it.

> Also, the sed invocation was changed quite a bit. Why?

too vague a question. :) The files in the chard are "yaml" and in examples they're "yml", we need to tell it to look into another directory as well. We could add a separate sed line for helm part.

> I think this could've been its own commit.

I don't think so but I don't have strong feelings about it so if you have, I can split it.

helm/habitat-operator/Chart.yaml
zeenix: after a bit of thinking and discussion on slack, I decided to keep the version in sync. It's easier to manage and the distributed charts won't cause any confusions.

zeenix: I mean the chart and the operator versions.

asymmetric:
@blixtra Since this PR introduces a Helm specific README, I think that info could go there.

zeenix:
@asymmetric way ahead of you. :) Info already added to that README. :)

Makefile
```diff
@@ -28,7 +28,13 @@ clean-test:
...
+		-exec sed -i.bak \
+		-e "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" \
```
asymmetric: What are the `-e` flags doing? Is it for executing multiple substitutions in one sed run?

helm/habitat-operator/templates/clusterrole.yaml
asymmetric: What should we do here? Or is this good?

helm/habitat-operator/templates/clusterrolebinding.yaml
asymmetric: Same, needs to be fixed?

Makefile
zeenix: -e is for specifying expressions but it can be (and is usually) omitted if you only have one expression.

helm/habitat-operator/templates/clusterrole.yaml
zeenix: I think so. I added the v1 case below.

helm/habitat-operator/templates/clusterrolebinding.yaml
zeenix: same answer here too. :)

Makefile
asymmetric: But `'.*.ya?ml$'?` is a regexp and regexps are 1337.

zeenix: No :)

helm/habitat-operator/README.md
```diff
@@ -0,0 +1,109 @@
+# habitat-operator
+
+Installs [habitat-operator](https://github.com/kinvolk/habitat-operator) to manage Habitat services in a Kubernetes cluster.
+
+## TL;DR;
+
+```console
+$ helm repo add habitat https://kinvolk.github.io/habitat-operator/helm/charts/stable/
+$ helm install habitat/habitat-operator
```
LiliC: Is this supposed to work already?

helm/habitat-operator/templates/clusterrole.yaml
LiliC: Yes, this is fine now!

helm/habitat-operator/README.md
zeenix: Yeah, I tested it here. If you have helm setup, please do try it out too.

helm/habitat-operator/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+```console
+$ helm install --name my-release habitat/habitat-operator -f values.yaml
+```
+
+> **Tip**: You can use the default [values.yaml](values.yaml)
+
+## Hosting
+
+The helm chart repository is managed on the [`gh-pages` branch](https://github.com/kinvolk/habitat-operator/tree/gh-pages) and instructions for repository management can be found [here](https://github.com/kinvolk/habitat-operator/tree/gh-pages/helm/charts).
```
LiliC: Nitpick warning: Capitalize Helm ;)

zeenix:
nitpick addressed. :)

PR #162, opened 2018-01-26
Use service name when forming path for user.toml

LiliC:
Due to recent changes, user now specifies in the `Service.Name` field the name of the service, and we do not take it from the habitat Name anymore. This should fix the issue we are seeing.

I am a bit confused why the tests passed, but it could also be why they were failing so often?

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -715,7 +715,7 @@ func (hc *HabitatController) newDeployment(h *habv1.Habitat) (*appsv1beta1.Deplo
 			Name: initialConfigFilename,
 			// The Habitat supervisor creates a directory for each service under /hab/svc/<servicename>.
 			// We need to place the user.toml file in there in order for it to be detected.
-			MountPath: fmt.Sprintf("/hab/svc/%s/%s", h.Name, userTOMLFile),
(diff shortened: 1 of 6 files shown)


LiliC:
@krnowak Good point! It's because the name is the same as the service name. https://github.com/kinvolk/habitat-operator/blob/master/test/e2e/resources/bind-config/habitat-postgresql.yml#L5 vs https://github.com/kinvolk/habitat-operator/blob/master/test/e2e/resources/bind-config/habitat-postgresql.yml#L10

Will change that to something "random", so we can find these problems easier next time!

krnowak:
This explanation is a too big mental shortcut for me to understand what's going on. :)

LiliC:
@krnowak So with the last changes the `Service.Name` should have been the one taken as the pkg name while formatting the path to the user.toml, but instead the `Name` was taken. But the `Name` in the test file stayed the same as the pkg name so the tests passed as they were formatting the user.toml path from the `Name` anyways. See line https://github.com/kinvolk/habitat-operator/blob/master/test/e2e/resources/bind-config/habitat-postgresql.yml#L5

LiliC:
@krnowak Adjusted the tests to some random name to avoid the same problem from happening, and if we mess something up with forming the path, we will then catch that.

PR #163, opened 2018-01-29
Add a section in main README about RBAC

krnowak:
It seems to confuse more and more people starting with the Habitat
Operator, so let's make it visible.

Also, renamed one manifest file to be more obvious about the contents
it has.


README.md
```diff
@@ -65,6 +65,20 @@ Habitat operator images are located [here](https://hub.docker.com/r/kinvolk/habi
 
 #### Deploying Habitat operator
 
+##### Cluster with RBAC enabled
+
+Since Kubernetes 1.8, RBAC is enabled by default, so for the operator to work, some rules needs to be deployed. To deploy them use the manifest file located under the examples directory:
```
schu: Thanks!

> Since Kubernetes 1.8, RBAC is enabled by default..

Can we not update `examples/habitat-operator-deployment.yml` to do the RBAC install (i.e. combine `rbac/rbac.yml` and `rbac/habitat-operator.yml`)?

schu: For example Flannel does that as well for a convenient single file install: https://github.com/coreos/flannel/blob/master/Documentation/kube-flannel.yml

LiliC: We do this in case the user has a different way of deploying the operator, e.g. just a `Pod`.

But we can just say to just run the following command which will deploy the operator as well the rules, this way it doesn't make a difference.

`kubectl create -f examples/rbac`

LiliC: > Since Kubernetes 1.8, RBAC is enabled by default, so for the operator to work, some rules needs to be deployed.

I am not 100% sure it's from 1.8.x, I do know it's stable since then but not certain it was enabled by default, at least not when I tried. My suggestion is below:

When running a Kubernetes cluster with [RBAC](https://kubernetes.io/docs/admin/authorization/rbac/) enabled, make sure to give Habitat operator the correct permissions to be able to create and monitor certain resources. To create the roles use the manifest file located under the examples directory:

`kubectl create -f examples/rbac`

krnowak:
Updated, how about now?

krnowak:
Issue about readme readability - https://github.com/kinvolk/habitat-operator/issues/168

PR #164, opened 2018-01-29
controller: Fix updating user.toml

krnowak:
Revert the SubPath usage and change the destination where we
bind-mount a directory. The new destination is supposed to contain
only the user.toml file, so hiding the previous contents of the
directory should not be a problem anymore. Not using SubPath works
around the atomic updates on bind-mounted file problems.

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -713,10 +713,9 @@ func (hc *HabitatController) newDeployment(h *habv1.Habitat) (*appsv1beta1.Deplo
 
 		secretVolumeMount := &apiv1.VolumeMount{
 			Name: initialConfigFilename,
-			// The Habitat supervisor creates a directory for each service under /hab/svc/<servicename>.
+			// The Habitat supervisor creates a directory for each service under /hab/user/<servicename>/config.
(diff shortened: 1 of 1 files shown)


pkg/controller/controller.go
```diff
@@ -713,10 +713,9 @@ func (hc *HabitatController) newDeployment(h *habv1.Habitat) (*appsv1beta1.Deplo
 
 		secretVolumeMount := &apiv1.VolumeMount{
 			Name: initialConfigFilename,
-			// The Habitat supervisor creates a directory for each service under /hab/svc/<servicename>.
+			// The Habitat supervisor creates a directory for each service under /hab/user/<servicename>/config.
```
LiliC: Really? Why was this decided to be user directory? 🤔

krnowak: Because I decided that. ;)

Basically, the contents of the svc directory should be for the service itself and should not be modified by user. So user.toml never really fit there.

We decided to go with a separate directory for user-editable stuff like user.toml.

LiliC: > Because I decided that. ;)

SGTM 👍

PR #169, opened 2018-01-29
Use numeric dependency versions for clarity

asymmetric:
Revisions are opaque, versions make it clear which version we're
running.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -298,6 +298,7 @@
     "storage/v1beta1"
   ]
   revision = "af4bc157c3a209798fc897f6d4aaaaeb6c2e0d6a"
+  version = "kubernetes-1.9.0"
 
 [[projects]]
   name = "k8s.io/apiextensions-apiserver"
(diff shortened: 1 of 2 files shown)


LiliC:
I would also bump to newer version of client-go, etc., since I changed this new releases were made.

asymmetric:
[v6](https://github.com/kubernetes/client-go/releases/tag/v6.0.0) is the latest stable AFAICT.

LiliC:
@asymmetric you can now take the k8s tags as well, the auto merging into that repo was fixed IIRC.

asymmetric:
I read [this](https://github.com/kubernetes/client-go#kubernetes-tags) and I have no idea what it means.

LiliC:
@asymmetric all of the client/lib repos come from the k8s main repos staging area. They have a tool that builds and "copies" the code over to the separate client/lib repos. There was an article published recently, you can have a look http://blog.kubernetes.io/2018/01/introducing-client-go-version-6.html

asymmetric:
Cool, thanks for the link. I still don't understand those `kubernetes-` tags, but I think that anyway upgrading can be done in another PR, if necessary.

asymmetric:
@LiliC In regards to updating to newer versions, I'd like to keep this PR as is, so that it's clear that we're not changing anything, just using a clearer naming.

I'll open a separate PR where I bump the minor versions.

krnowak:
LFAD if an issue or a PR about upgrading is filed too. Or is #173 about it too?

asymmetric:
I've created #175 for that.

PR #170, opened 2018-01-29
Use pflag in tests

### LINKED ISSUE #78: Switch to pflag in test
Tried using [pflag](https://github.com/spf13/pflag) but it did not work, because flag registering/parsing seem to get overridden. Maybe it's because other parts of the code/dependencies parse things through init methods, but it needs some further investigation.

My attempt:
```
type testFlag struct {
	image      string
	kubeconfig string
	externalIP string
}
....
flags := flag.NewFlagSet(os.Args[0], flag.ContinueOnError)

flags.StringVar(&tf.image, "image", "", "habitat operator image, 'kinvolk/habitat-operator'")
flags.StringVar(&tf.kubeconfig, "kubeconfig", "", "path to kube config file")
flags.StringVar(&tf.externalIP, "ip", "", "external ip, eg. minikube ip")

flags.Parse(os.Args[2:]) // As the previous flags are test related flags.
```

asymmetric:
The pflag package is used elsewhere because it allows us to set
shorthands (e.g. -v for --verbose).

We now use in this test file as well for consistency, although we're not
using any of its differentiating features.

Closes #78.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -15,12 +15,12 @@
 package e2e
 
 import (
-	"flag"
 	"fmt"
 	"os"
 	"testing"
 
 	operatorFramework "github.com/kinvolk/habitat-operator/test/e2e/framework"
+	flag "github.com/spf13/pflag"
 )
 
 var framework *operatorFramework.Framework


LiliC:
@asymmetric your tests seem to be failing and it seems to be the same issue as I was experiencing.

krnowak:
Can't see anything obviously wrong with this change, but it would be nice to rerun the tests so they are green.

asymmetric:
Just rerunning won't fix this, there's something else going wrong but the issue is not very important. Feel free to ignore for now.

PR #85, opened 2017-08-30
Rename CRD from ServiceGroup to Habitat

asymmetric:
This removes the amibguity between the CRD name and the habitat concept
of a service group.

The short name becomes `hab`.

I'm not 100% about this rename. For example, the full CRD name is now `habitats.habitat.sh`.

Closes #85.

Issue #39, opened 2017-08-07
Rename CRD ServiceGroup


LiliC:
Current name `ServiceGroup` conflicts with concept of a `group` in Habitat, as well as concept of a `Service` in Kubernetes, we need to come up with a better name for CRD.

List of ideas:
- [HabitatService](https://github.com/kinvolk/habitat-operator/pull/30)
- Habitat

asymmetric:
What changed since we decided to close #30?

LiliC:
We decided to close that PR as we couldn't come up with anything better then `ServiceGroup`, but we agreed that `ServiceGroup` is not the best name from what I remember? 

In one of our last meetings I believe we sort of agreed on `Habitat`, but I wanted your input before changing everything to that, so I opened the issue to not forget and also to discuss/collect ideas.

asymmetric:
A pro of using "Habitat" would be that we would follow the logic of other operators:

* PrometheusOperator -> Prometheus
* ReplicaSetController -> ReplicaSet
* ...

LiliC:
Closed by https://github.com/kinvolk/habitat-operator/pull/85

PR #172, opened 2018-01-30
Update user.toml path and example images

### LINKED ISSUE #165: Limit the number of docker images we use in the examples
Currently we use 5 of them:

- kinvolk/postgresql-hab
- kinvolk/bindgo-hab
- kinvolk/nodejs-hab
- kinvolk/consul-hab
- kinvolk/nginx-hab

The first two are used in tests. I wonder if we could replace them with fewer images and smaller. (postgresql image is 600 mb).

That should make it a bit less painful when updating the images to use newer habitat version.

asymmetric:
This PR updates the images used in examples. The new images feature an updated supervisor, which looks for the `user.toml` file in a [new location](https://github.com/habitat-sh/habitat/pull/3814).

Also, this PR changes the example to use redis whenever possible, to reduce maintenance overhead.

Closes #164.
Closes #165.


LiliC:
I would like to see a test for configuration updates as part of this PR, wdyt?

asymmetric:
@LiliC what kind of tests were you thinking of?

LiliC:
@asymmetric Similar as we already have for initial config, but this to make sure config updates work and will always work despite the fact we change something.

asymmetric:
I'd keep that for a separate PR. I don't see a very strong correlation between this PR and tests for `user.toml` updates.

Also, this PR is blocking the release. Or do you mean that we should have the tests before we release?

Either way, I'd keep the two things separate.

examples/config/habitat.yml
```diff
@@ -4,10 +4,9 @@ metadata:
   name: user-toml
 type: Opaque
 data:
-  # Each item needs to be encoded in base64, as Kubernetes
-  # expects that encoding.
-  # Plain text content: protected-mode = "no"
-  user.toml: cHJvdGVjdGVkLW1vZGUgPSAibm8i
+  # Each item needs to be base64-encoded in base64.
```
krnowak: The reason why I went with the protected-mode, instead of port number was that we can see a different message when visiting <cluster_ip>:30001. With protected mode set by default to "yes", we get a message about Redis being locked down. With our user.toml override, we could see something else.

But in the end, changing the port number is also fine - you documented that below in the Service part.

examples/config/README.md
```diff
@@ -1,6 +1,6 @@
 # Initial configuration
 
-This example demostrates how initial configuration works with the Habitat operator. With the manifest file we deploy a Redis Habitat service.
+This example demonstrates how initial configuration works with the Habitat operator. With the manifest file we deploy a Redis Habitat service.
```
krnowak: I'd be against fixing the typo - the word with the typo sounds like a Greek name and we need some greek reference, because Kubernetes.

krnowak:
Filed an issue for adding a test for config updates, so we won't forget that.

examples/bind-config/habitat.yml
```diff
@@ -6,18 +6,18 @@ type: Opaque
 data:
   # Each item needs to be encoded in base64, as Kubernetes expects that encoding.
   # Plain text content of the secret: "port = 4444"
-  # This overrides the port set in the postgresql Habitat service.
+  # This overrides the port set in the Redis Habitat service.
   user.toml: cG9ydCA9IDQ0NDQ=
 ---
 apiVersion: habitat.sh/v1
-kind: Habitat 
+kind: Habitat
```
zeenix: irrelevant change. Separate patch for fixing formatting please

examples/bind/README.md
```diff
@@ -10,6 +10,6 @@ After the Habitat operator is up and running, execute the following command from
 kubectl create -f examples/bind/habitat.yml
 ```
 
-This will deploy two `Habitat`s, a simple HTTP server written in Go that will be bound to a PostgreSQL database. The Go server will display the port number the database listens on.
+This will deploy two `Habitat`s, a simple HTTP server written in Go that will be bound to a Redis database. The Go server will display the port number the database listens on.
```
zeenix: Do we really call the services or apps as "habitats"?

examples/bind-config/README.md
```diff
@@ -10,6 +10,12 @@ After the Habitat operator is up and running, execute the following command from
 kubectl create -f examples/bind-config/habitat.yml
 ```
 
-This will deploy two `Habitat`s, a simple HTTP server written in Go that will be bound to a Redis database. The Go server will display the database port number that was overriden by the initial configuration.
+This will deploy two `Habitat`s, a simple HTTP server written in Go that will be
+bound to a Redis instance. By default, the Redis database instance would [listen
+on port
+6379](https://github.com/habitat-sh/core-plans/blob/7bc934c31e92c959aea0444671900c57c23d5265/redis/default.toml#L3),
+but we change this with the configuration stored in the `user-toml`.
```
zeenix: in the commit log it's "user.toml", but here it's with a dash.

examples/bind/README.md
asymmetric: Yeah, the naming is a bit weird. In Habitat they're services, but the CRD is Habitat, so same as we say ConfigMaps, I think we're justified to say Habitats.

examples/bind-config/README.md
asymmetric: It's with a dash because what I'm pointing at here is the Kubernetes Secret named `user-toml`, rather than the filename under which the Secret is mounted.

pkg/controller/controller.go
```diff
@@ -715,7 +715,7 @@ func (hc *HabitatController) newDeployment(h *habv1.Habitat) (*appsv1beta1.Deplo
 			Name: initialConfigFilename,
 			// The Habitat supervisor creates a directory for each service under /hab/svc/<servicename>.
 			// We need to place the user.toml file in there in order for it to be detected.
-			MountPath: fmt.Sprintf("/hab/svc/%s/%s", h.Spec.Service.Name, userTOMLFile),
+			MountPath: fmt.Sprintf("/hab/user/%s/config/%s", h.Spec.Service.Name, userTOMLFile),
 			SubPath:   userTOMLFile,
```
krnowak: We need to get rid of this line.

examples/bind/README.md
zeenix: Not sure I agree and it sounds strange but OK. :)

examples/config/README.md
zeenix: I'm actually against it but for a very different reason: it's not relevant to this patch. :)

examples/encrypted/README.md
```diff
@@ -34,9 +34,3 @@ kubectl create -f examples/encrypted/habitat.yml
 The Habitat operator does not delete the Secret on Habitat deletion. This is
 because the user might want to re-use the secret across multiple
 `Habitat`s and `Habitat` lifecycles.
-
-To delete the Secret simply run:
-
-```
-  kubectl delete secret your-secret-name
-```
```
zeenix: but we are not teaching k8s, just giving them a direct command to use so they don't have to look it up just for this.

asymmetric:
@zeenix Originally I've learned of stuttering as a naming anti-pattern first encountered [here](https://golang.org/doc/effective_go.html), it basically means that names should not repeat: `node.Node` would be stutter, `buffer.BufferAt`.

In this case, a Service whose name ends in `-service`, and stuff like that.

[Some more info](https://github.com/golang/go/wiki/CodeReviewComments).

examples/encrypted/README.md
asymmetric: Yeah but this is basic stuff, and it's also orthogonal: create/delete work across all resources. If you don't know how to do CRUD operations on a resource, I don't think you should delve into operators yet..

examples/bind-config/habitat.yml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: Secret
 metadata:
-  name: user-toml-secret
+  name: user-toml
```
zeenix: Thanks for explaining what "stutter" is but i don't see any repetition here in the name.

examples/encrypted/README.md
asymmetric: But I mean, I can add it back.. I just think it's useless, as virtually 100% of people reading this will know how to do it.

examples/bind-config/habitat.yml
asymmetric: > In this case, a Service whose name ends in -service, and stuff like that.

asymmetric: `kubectl get secret` returning `user-toml-secret` is stutter IMO. Of course it's a Secret. It doesn't help to have `-secret` there, because that object will only be returned when you ask for Secrets.

zeenix: I agree with the change. I was just being nitpicky about it not fitting the "stutter" title.

examples/config/README.md
```diff
@@ -9,8 +9,14 @@ After the Habitat operator is up and running, execute the following command from
 
 `kubectl create -f examples/config/habitat.yml`
 
-This will create a [Kubernetes Secret](https://kubernetes.io/docs/concepts/configuration/secret/) with the configurations and a simple Node.js application that will display a msg. When running on minikube, it can be accessed under port `30001` of the minikube VM. `minikube ip` can be used to retrieve the IP.
-Initially our app is configured to display the msg `"Hello world."`. Because we override this with the Secret we just created, our app will instead display `Hello from our Habitat-Operator!`.
+This will create a [Kubernetes Secret](https://kubernetes.io/docs/concepts/configuration/secret/) with the configurations and a Redis database.
+Initially the Redis database is configured to be in protected mode. Because we override this with the Secret we just created, our db will not be in this mode anymore.
```
krnowak: Drop this line, please - you decided to change the port instead of the protected mode.

examples/bind/README.md
asymmetric: I agree that it sounds strange, but this was the result of a long discussion and we couldn't come up with a better name.

See [this](https://github.com/kinvolk/habitat-operator/pull/85) and [this](https://github.com/kinvolk/habitat-operator/issues/39).

PR #178, opened 2018-02-01
Automate most of Helm chart generation

zeenix:
Automation means emancipation of the masses!

### DIFF
--- a/helm/charts/README.md
+++ b/helm/charts/README.md
@@ -2,14 +2,10 @@
 
 This directory hosts all the helm charts related to the habitat-operator.
 
-To add chart for a new release of habitat-operator:
+To add chart for a new release of habitat-operator run the following from the `helm/chart` directory:
 
 ```console
-$ cd stable # assuming you're already in helm/charts directory
-$ helm package ../../habitat-operator
-$ helm repo index . --url https://kinvolk.github.io/habitat-operator/helm/charts/stable/ --merge index.yaml
-$ git add index.yaml habitat-operator-RELEASE.tgz
-$ git commit -m 'Add helm chart for RELEASE'
+$ ./chart-for-latest.sh
 $ git push origin gh-pages
 ```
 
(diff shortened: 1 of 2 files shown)


helm/charts/Makefile
```diff
@@ -0,0 +1,6 @@
+all:
```
LiliC: Can you add this maybe to a new make target to the main Makefile. Called something like `make release` for example?

So we in the end can do something like `make release` and that calls first `update-version` and then this `make all`. WDYT? (We automate things even more this way ;))

helm/charts/README.md
```diff
@@ -5,11 +5,7 @@ This directory hosts all the helm charts related to the habitat-operator.
 To add chart for a new release of habitat-operator:
 
 ```console
-$ cd stable # assuming you're already in helm/charts directory
-$ helm package ../../habitat-operator
-$ helm repo index . --url https://kinvolk.github.io/habitat-operator/helm/charts/stable/ --merge index.yaml
-$ git add index.yaml habitat-operator-RELEASE.tgz
-$ git commit -m 'Add helm chart for RELEASE'
+$ make                      # assuming you're already in helm/charts directory
```
LiliC: Maybe instead of this just rewrite the above msg: `To add the chart for a new release of the habitat-operator run the following from the `helm/chart` directory:`

(Unless you are doing the first comment, then this needs to be changed anyways. :)

helm/charts/Makefile
LiliC: NVM. Just figured out it's not going into master :(

helm/charts/README.md
zeenix: done.

zeenix:
Pushed the change to turn Makefile into a shell script.

zeenix:
> gh-pages branch being a master branch + some extras… looks terrible.

AFAICT that's the recommend way from helm folks:

> In such setup you can use master branch to store your charts code, and gh-pages branch as charts repository

 https://github.com/kubernetes/helm/blob/master/docs/chart_repository.md#github-pages-example

helm/charts/chart-for-latest.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
```
krnowak: Please add `set -e` to make the script fail on the first failed command here.

zeenix: done

PR #179, opened 2018-02-01
Release 0.5.0

zeenix:
I created the changelog for 0.5.0 from git log. Please have a look and be gentle, it's my first time. :D


CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
+## [v0.5.0](https://github.com/kinvolk/habitat-operator/tree/v0.5.0) (1-2-2018)
+[Full changelog](https://github.com/kinvolk/habitat-operator/compare/v0.4.0...v0.5.0)
+
+### Features & Enhancements
+
+- Switch to e2e testing against RBAC enabled cluster and refactor tests [#151](https://github.com/kinvolk/habitat-operator/pull/151)
+- Add Name field [#155](https://github.com/kinvolk/habitat-operator/pull/155)
+- Clean up unused imports [#160](https://github.com/kinvolk/habitat-operator/pull/160)
+- Use service name when forming path for user.toml [#162](https://github.com/kinvolk/habitat-operator/pull/162)
+- Add a section in main README about RBAC [#163](https://github.com/kinvolk/habitat-operator/pull/163)
```
LiliC: This doesn't need to be in the release notes.

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
 # Habitat operator CHANGELOG
 
+## [v0.5.0](https://github.com/kinvolk/habitat-operator/tree/v0.5.0) (1-2-2018)
+[Full changelog](https://github.com/kinvolk/habitat-operator/compare/v0.4.0...v0.5.0)
+
+### Features & Enhancements
+
+- Switch to e2e testing against RBAC enabled cluster and refactor tests [#151](https://github.com/kinvolk/habitat-operator/pull/151)
+- Add Name field [#155](https://github.com/kinvolk/habitat-operator/pull/155)
+- Clean up unused imports [#160](https://github.com/kinvolk/habitat-operator/pull/160)
```
LiliC: This is also irrelevant IMO.

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
+- Use service name when forming path for user.toml [#162](https://github.com/kinvolk/habitat-operator/pull/162)
+- Add a section in main README about RBAC [#163](https://github.com/kinvolk/habitat-operator/pull/163)
+- Downgrade API to v1beta1 [#167](https://github.com/kinvolk/habitat-operator/pull/167)
+- Use numeric dependency versions for clarity [#169](https://github.com/kinvolk/habitat-operator/pull/169)
```
LiliC: Irrelevant as well, does not need to mentioned.

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
 
+## [v0.5.0](https://github.com/kinvolk/habitat-operator/tree/v0.5.0) (1-2-2018)
+[Full changelog](https://github.com/kinvolk/habitat-operator/compare/v0.4.0...v0.5.0)
+
+### Features & Enhancements
+
+- Switch to e2e testing against RBAC enabled cluster and refactor tests [#151](https://github.com/kinvolk/habitat-operator/pull/151)
+- Add Name field [#155](https://github.com/kinvolk/habitat-operator/pull/155)
+- Clean up unused imports [#160](https://github.com/kinvolk/habitat-operator/pull/160)
+- Use service name when forming path for user.toml [#162](https://github.com/kinvolk/habitat-operator/pull/162)
```
krnowak: I think I'd drop this too.

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
+- Add a section in main README about RBAC [#163](https://github.com/kinvolk/habitat-operator/pull/163)
+- Downgrade API to v1beta1 [#167](https://github.com/kinvolk/habitat-operator/pull/167)
```
krnowak: This could be in the breaking changes section.

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
+- Use numeric dependency versions for clarity [#169](https://github.com/kinvolk/habitat-operator/pull/169)
+- Update user.toml path and example images [#172](https://github.com/kinvolk/habitat-operator/pull/172)
+- Use cache for ConfigMaps [#157](https://github.com/kinvolk/habitat-operator/pull/157)
+- Avoid installing dependencies when building image [#142](https://github.com/kinvolk/habitat-operator/pull/142)
```
krnowak: Drop this one.

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
 # Habitat operator CHANGELOG
 
+## [v0.5.0](https://github.com/kinvolk/habitat-operator/tree/v0.5.0) (1-2-2018)
+[Full changelog](https://github.com/kinvolk/habitat-operator/compare/v0.4.0...v0.5.0)
+
+### Features & Enhancements
+
+- Switch to e2e testing against RBAC enabled cluster and refactor tests [#151](https://github.com/kinvolk/habitat-operator/pull/151)
+- Add Name field [#155](https://github.com/kinvolk/habitat-operator/pull/155)
```
krnowak: This is a breaking change, so would be good if it landed in the breaking changes section.

CHANGELOG.md
```diff
@@ -1,5 +1,22 @@
...
+- Clean up unused imports [#160](https://github.com/kinvolk/habitat-operator/pull/160)
+- Use service name when forming path for user.toml [#162](https://github.com/kinvolk/habitat-operator/pull/162)
+- Add a section in main README about RBAC [#163](https://github.com/kinvolk/habitat-operator/pull/163)
+- Downgrade API to v1beta1 [#167](https://github.com/kinvolk/habitat-operator/pull/167)
+- Use numeric dependency versions for clarity [#169](https://github.com/kinvolk/habitat-operator/pull/169)
+- Update user.toml path and example images [#172](https://github.com/kinvolk/habitat-operator/pull/172)
```
krnowak: This probably could be reworded: "Make configuration updates through Secrets work".

LiliC: Considering the examples, docs or tests for configuration updates were never added, I wouldn't say that. This just updates the path IMO.

CHANGELOG.md
LiliC: It was a bug fix?

krnowak: A bug fix that did not exist in the previous release, but was introduced with the new Name field.

CHANGELOG.md
krnowak: Fair enough.

CHANGELOG.md
zeenix: AFAICT this is not breaking change: existing habitat manifests will continue to work like before. If not, the title and description of the PR is very misleading.

zeenix: I am in general a bit confused about this change. What's wrong with making use of metadata name?

zeenix: I think I addressed all other comments except for this one for reasons mentioned above.

zeenix: Thanks. Would have been nice for operator to just use metadata name if service name wasn't provided and warn about it in the log? I-e for service name to be optional in case user won't care and want to use the same name for both.

Please feel free to take over this PR now. :)

LiliC: @zeenix That PR was already merged, feel free to open an issue or discuss with @asymmetric about it, if you want to resolve that before we can even postpone this release?

CHANGELOG.md
```diff
@@ -1,5 +1,20 @@
 # Habitat operator CHANGELOG
 
+## [v0.5.0](https://github.com/kinvolk/habitat-operator/tree/v0.5.0) (1-2-2018)
+[Full changelog](https://github.com/kinvolk/habitat-operator/compare/v0.4.0...v0.5.0)
+
+### Breaking changes
+
+* API downgrade from "v1" to "v1beta1" to better reflect the API instability [#167](https://github.com/kinvolk/habitat-operator/pull/167)
```
LiliC: This could be a bit more helpful. For example: 

> The habitat resource API version is now "v1beta1", check the examples section to see how to make the new changes.

CHANGELOG.md
zeenix: right, i'm discussing with him now.

asymmetric: I think that would complicate the code and also make it less clear for users in a way. Now each field has a well-defined purpose. If we do what you're asking, the Metadata.Name field will have two purposes, depending on whether another field is defined.

zeenix: (slightly) complicating the code for making user's life easier, is a nobrainer for me. UX should be the highest priority in every software project (TM).

I don't agree with this being complicated to users. We should document that if service name is optional and if not provided, metadata name will be used.

asymmetric: Not sure it's easier for users. Opinions @krnowak @LiliC?

zeenix: In the simple case where user don't care about names being the same, I'd say it is.

LiliC: My opinion is that we are in the beta and breaking changes can occur in the API. We have that written in the README. As for the k8s exporter it wasn't even included in Habitat yet, we can change things there as well. I am with @asymmetric on this one, this is the best approach one field does one thing, we make too many assumptions otherwise.

zeenix: sure if it makes sense to break it, go for it but in this case it makes more sense not to break it instead. I don't know, to me this is a very much nobrainer so I don't understand the resistance to my idea but I won't lose any sleep on this.

CHANGELOG.md
zeenix: PTALA.

zeenix:
PTALA

LiliC:
@zeenix #183 is missing ;)

zeenix:
> @zeenix #183 is missing ;)

Not anymore. :)

CHANGELOG.md
```diff
@@ -1,5 +1,23 @@
...
+
+- API downgrade from "v1" to "v1beta1" to better reflect the API instability [#167](https://github.com/kinvolk/habitat-operator/pull/167)
+- Add Name field [#155](https://github.com/kinvolk/habitat-operator/pull/155)
+
+Please refer to examples for how to adapt existing manifests.
+
+### Features & Enhancements
+
+- Switch to e2e testing against RBAC enabled cluster and refactor tests [#151](https://github.com/kinvolk/habitat-operator/pull/151)
+- E2E test for user.toml update [#183](https://github.com/kinvolk/habitat-operator/pull/183)
```
LiliC: I would mention something like configuration updates in this release notes. Not sure where the right place is. WDYT @asymmetric

asymmetric: Yeah, that's a good point, I think this is the first time that a supervisor with support for `user.toml` updates is shipped. Also we adjusted the path we place the `user.toml` in.

You can write something like:

* Added support for configuration updates via `user.toml`
* Updated mount path of `user.toml` config file (and maybe here you can reference the PR where they changed that - @krnowak what was it?)
* Updated supervisor in examples to version x.y.z

krnowak: https://github.com/habitat-sh/habitat/pull/3814

zeenix: > Added support for configuration updates via user.toml

Added or fixed?

LiliC: Added! We did not have config updates, but initial configuration before.

zeenix: what's 'x.y.z'?

asymmetric: You mean, which version of the supervisor are we running in examples? @krnowak 

Don't know if we need to mention it in the changelogs though.

LiliC: Good point! Forgot if we changed that? It's currently `Habitat >= 0.36.0`. Is that correct @krnowak. If not we should update before merging this ;)

zeenix: @asymmetric you suggested to mention it. :)

krnowak: @asymmetric : The alternative path for user.toml was introduced in 0.39. We are running version 0.52 in our examples.

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+### Breaking changes
+
+- API downgrade from "v1" to "v1beta1" to better reflect the API instability [#167](https://github.com/kinvolk/habitat-operator/pull/167)
+- Add Name field [#155](https://github.com/kinvolk/habitat-operator/pull/155)
+
+Please refer to examples for how to adapt existing manifests.
+
+### Features & Enhancements
+
+- Switch to e2e testing against RBAC enabled cluster and refactor tests [#151](https://github.com/kinvolk/habitat-operator/pull/151)
```
asymmetric: I don't think users are interested in us adding tests

CHANGELOG.md
```diff
@@ -1,5 +1,24 @@
...
+- Updated mount path of user.toml config file as per [Habitat change](https://github.com/habitat-sh/habitat/pull/3814)
+- E2E test for user.toml update [#183](https://github.com/kinvolk/habitat-operator/pull/183)
```
asymmetric: Ditto.

asymmetric:
PTALA.

PR #183, opened 2018-02-07
E2E test for user.toml update

asymmetric:
Doesn't work:

> operator_test.go:160: Configuration update did not go through. Expected: "hello from port: 6333", got: "hello from port: 4444"

I tried increasing the sleep time but that didn't help.

### DIFF
--- a/examples/config/README.md
+++ b/examples/config/README.md
@@ -1,4 +1,4 @@
-# Initial configuration
+# User configuration
 
 This examples demonstrates how to leverage the `user.toml` configuration
 mechanism for Habitat Services within the operator.
--- a/examples/config/README.md
+++ b/examples/config/README.md
@@ -14,8 +14,8 @@ This will create a [Kubernetes Secret](https://kubernetes.io/docs/concepts/confi
 By default, Redis listens on port 6379, but we change this to 6999 by mounting a
 Secret as a file under `/hab/user/redis/config/user.toml` inside the Pod.
 
-The web app is listening on port `30001`. When running on minikube, its IP can
-be retrieved with `minikube ip`.
+You can see this is the case by accessing the web app on port `30001`. When
+running on minikube, its IP can be retrieved with `minikube ip`.
 
 ## Deletion
 
--- a/examples/config/README.md
+++ b/examples/config/README.md
@@ -23,6 +23,23 @@ Secret as a file under `/hab/user/redis/config/user.toml` inside the Pod.
 You can see this is the case by accessing the web app on port `30001`. When
 running on minikube, its IP can be retrieved with `minikube ip`.
 
+## Configuration updates
+
+In Kubernetes, a file backed by a Secret is kept in sync with the Secret: when
+the Secret changes, the file does as well (after a short delay).
+
+You can try this yourself by editing the Secret:
+
+    echo "port = 6160" | base64
+    kubectl edit secret user-toml
+
+As you type the last command, your default editor will be invoked. You should
+then paste the new base64-encoded string under `data.user-toml`, and exit the
+editor.
+
+After a short delay, you should be able to see port 6160 being used by the Redis
+service.
+
 ## Deletion
 
 The Habitat operator does not delete the Secret on Habitat deletion, as it is not managed by the Habitat operator.
--- a/examples/config/README.md
+++ b/examples/config/README.md
@@ -13,7 +13,7 @@ NOTE: Adding secret configuration to the `default.toml` is discouraged, as it wi
 
 After the Habitat operator is up and running, execute the following command from the root of this repository:
 
-`kubectl create -f examples/config/habitat.yml`
+    kubectl create -f examples/config/habitat.yml
 
 This will create a [Kubernetes Secret](https://kubernetes.io/docs/concepts/configuration/secret/) with the configurations and a Redis database.
 
@@ -45,6 +45,4 @@ service.
 The Habitat operator does not delete the Secret on Habitat deletion, as it is not managed by the Habitat operator.
 To manually delete the Secret simply run:
 
-```
-kubectl delete service user-toml-secret
-```
+    kubectl delete service user-toml-secret
--- a/examples/config/README.md
+++ b/examples/config/README.md
@@ -1,6 +1,12 @@
 # Initial configuration
 
-This example demonstrates how initial configuration works with the Habitat operator. With the manifest file we deploy a Redis Habitat service.
+This examples demonstrates how to leverage the `user.toml` configuration
+mechanism for Habitat Services within the operator.
+
+When a `user.toml` file is located at `/hab/user/$servicename/config/`, it is
+automatically loaded by the supervisor. We leverage Kubernetes Secrets to mount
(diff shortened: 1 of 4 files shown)


test/e2e/operator_test.go
```diff
@@ -122,8 +123,45 @@ func TestFunction(t *testing.T) {
 		t.Fatalf("Habitat Service msg does not match one in default.toml. Expected: \"%s\", got: \"%s\"", expectedMsg, actualMsg)
 	}
 
+	// Update secret.
+	newPort := "6333"
+	encoded := make([]byte, base64.StdEncoding.EncodedLen(len(newPort)))
```
LiliC: The `Secret` does not need to be encoded as you are overriding it directly via the API. If you were to use it from a json file yes, there it needs to be encoded but here it can be in plain text.

asymmetric: Another mistake I was making is that I was setting `Data["user.toml"]` to just 6333, instead of `port = 6333`.

test/e2e/operator_test.go
```diff
@@ -122,8 +122,43 @@ func TestFunction(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Update the message set in the config of the kinvolk/bindgo-hab Go Habitat Service.
+	expectedMsg = fmt.Sprintf("hello from port: %v", 6333)
+	actualMsg = string(bodyBytes)
+	// actualMsg can contain whitespace and newlines or different formatting,
+	// the only thing we need to check is it contains the expectedMsg.
+	if !strings.Contains(actualMsg, expectedMsg) {
```
LiliC: As this is now duplicated, we should refactor this a bit, wdyt?

examples/config/README.md
```diff
@@ -14,8 +14,8 @@ This will create a [Kubernetes Secret](https://kubernetes.io/docs/concepts/confi
 By default, Redis listens on port 6379, but we change this to 6999 by mounting a
 Secret as a file under `/hab/user/redis/config/user.toml` inside the Pod.
 
-The web app is listening on port `30001`. When running on minikube, its IP can
```
LiliC: There is no guide on how to do config updates. So think that would be useful to be added here.

test/e2e/operator_test.go
```diff
@@ -122,8 +122,43 @@ func TestFunction(t *testing.T) {
 		t.Fatalf("Habitat Service msg does not match one in default.toml. Expected: \"%s\", got: \"%s\"", expectedMsg, actualMsg)
 	}
 
+	// Update secret.
+	newPort := "port = 6333"
+
+	sec.Data["user.toml"] = []byte(newPort)
```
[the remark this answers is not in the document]
asymmetric: Hmm.. I can see what you mean, but OTOH we are not testing Kubernetes itself. The operator doesn't manage `Secret` updates - Kubernetes does. So if Kubernetes works correctly, both approaches are identical - and it's outside of the scope of our tests to prove that Kubernetes works correctly. WDYT?

test/e2e/operator_test.go
```diff
@@ -122,8 +122,43 @@ func TestFunction(t *testing.T) {
...
+	if _, err = framework.KubeClient.CoreV1().Secrets(utils.TestNs).Update(sec); err != nil {
+		t.Fatalf("Could not update Secret: \"%s\"", err)
+	}
+
+	// Wait for SecretVolume to be updated.
+	time.Sleep(defaultWaitTime)
```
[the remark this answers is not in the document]
asymmetric: It's the same delay that used to be known as `waitForPorts`. In both cases, we could be experiencing test brittleness, and in that case I'd increase the delay.

But since we haven't yet, I'd leave this as is. Basically, fix it only if it breaks.

test/e2e/operator_test.go
asymmetric: Which part would you refactor?

LiliC: [This](https://github.com/kinvolk/habitat-operator/blob/fb4d1c39fadcb6559ae40e032e297f6964472857/test/e2e/operator_test.go#L99-L123) and [this](https://github.com/kinvolk/habitat-operator/blob/fb4d1c39fadcb6559ae40e032e297f6964472857/test/e2e/operator_test.go#L136-L159) seem to be very much duplicated. Could be extracted to another helper function, wdyt?

test/e2e/operator_test.go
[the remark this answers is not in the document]
LiliC: sgtm

asymmetric:
PTALA.

examples/config/README.md
```diff
@@ -1,27 +1,48 @@
...
-be retrieved with `minikube ip`.
+You can see this is the case by accessing the web app on port `30001`. When
+running on minikube, its IP can be retrieved with `minikube ip`.
+
+## Configuration updates
+
+In Kubernetes, a file backed by a Secret is kept in sync with the Secret: when
+the Secret changes, the file does as well (after a short delay).
```
LiliC: Maybe explain the delay has nothing to do with the operator, but k8s itself?

examples/config/README.md
```diff
@@ -1,27 +1,48 @@
...
+You can try this yourself by editing the Secret:
+
+    echo "port = 6160" | base64
+    kubectl edit secret user-toml
+
+As you type the last command, your default editor will be invoked. You should
+then paste the new base64-encoded string under `data.user-toml`, and exit the
+editor.
+
+After a short delay, you should be able to see port 6160 being used by the Redis
```
LiliC: You really hate the delay, don't you :D

test/e2e/framework/helpers.go
```diff
@@ -247,3 +250,22 @@ func pathToOSFile(relativePath string) (*os.File, error) {
 
 	return manifest, nil
 }
+
+func QueryService(url string) (string, error) {
```
LiliC: Function documentation comment missing? :)

test/e2e/operator_test.go
```diff
@@ -29,101 +27,118 @@ import (
...
+func TestBind(t *testing.T) {
 	// Get Habitat object from Habitat go example.
-	habitatGo, err := utils.ConvertHabitat("resources/bind-config/webapp.yml")
+	wapp, err := utils.ConvertHabitat("resources/bind-config/webapp.yml")
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	if err := framework.CreateHabitat(habitatGo); err != nil {
+	if err := framework.CreateHabitat(wapp); err != nil {
```
LiliC: wapp? :D

examples/config/README.md
asymmetric: Oh yes.

examples/config/README.md
asymmetric: I think it's kinda clear from the sentence that we're explaining how Kubernetes works, not the Operator. Or?

LiliC: Btw this is exactly what I was saying when you forcefully break lines in files. :)

Issue #181, opened 2018-02-06
RFE: support env and volumeMounts in HabitatSpec


jeremymv2:
This is a request for two enhancements:

 - [ ] It would be nice to be able to support habitat clusters with persistent disk requirements, such that `volumeMounts` could be defined and taken into account when the Deployment is created in kubernetes. This would enable users to create rings with services that mount a filesystem backed datastore allowing dependent services to query the ring for such things as the db service's ip/port.
 - [x] Additionally, being able to pass supervisor configuration via HAB_SVC_NAME environment variables to the containers would, I believe, be more straightforward and transparent then utilizing base64encoded `toml` implemented as a `Secret`. **[done in #184]**

asymmetric:
Hi @jeremymv2!

We're investigating approaches to provide persistent storage. One approach is to use [`StatefulSets`](44) instead of Deployments, and add support for `volumeMounts` in the Habitat object. 

Another approach we're exploring is to relegate persistence to services outside the cluster: a stateful service running outside Kubernetes forming a ring with a stateless one inside it.

Your usecase is helpful in that it allows us to gauge interest and needs, so please feel free to add other details if they come to mind.

The environment variables part is easier to figure out/implement, I've added it to my to-do list!

jeremymv2:
@asymmetric for a use-case, this is something I'd love to be able to migrate to the CRD habitat-operator approach.

https://github.com/jeremymv2/launch-chef-in-kubernetes/blob/master/chef-server-pod.yml

jeremymv2:
@asymmetric the challenge is that my Chef Server use-case above has a mix of stateful/stateless services that all need to join a habitat ring.

Issue #182, opened 2018-02-07
Define rationale for Habitat type structure


asymmetric:
It's not totally clear at the moment why we put keys where we put them in the `Habitat` type.

* Is there a clear rationale?
* Would it make more sense to flatten it by removing the `Service` subtype?

I say this because I'm having trouble justifying why `configSecretName` should be under `Service`, rather than under `Habitat` directly.

Also, every time we have to add a new field, we should clearly know where.

---

I thought a clear rationale was: every field that represents a flag passed to the supervisor should go under `Service`. So for example: `--name`, `--topology`, `--bind`.

If that's true, than `configSecretName` should be move up.

But I wonder if such a rationale makes sense from the point of view of a user. It is not based on a logical grouping, rather it's based on inheritance (of the flags used by an existing tool).

asymmetric:
I think we probably don't need to have a `service` key at all, i.e. we can just one level of depth.

asymmetric:
@zeenix WDYT? Flattening the type layout will be a breaking change, and you seem to have strong feelings towards backwards compatibilty.

LiliC:
@asymmetric Can you paste a sample manifest file of your idea for better clarity. 

Also as a side note according to the API version we are in beta not alpha, so these changes should have been done before that.

asymmetric:
Basicaly,

```yaml
apiVersion: habitat.sh/v1beta1
kind: Habitat
metadata:
  name: example-configured-habitat
spec:
  image: kinvolk/redis-hab
  count: 1
  service-name: db
  topology: standalone
  group: redisdb
  configSecretName: user-toml
```

zeenix:
well I'm against needless API breakage but I'm not sure it's the case in here. However, we should try our best to design this well right now if we're breaking again and also consult Chef folks about breaking this API and how it affects the k8s and helm exporters.

asymmetric:
After thinking about this some more, I think the current structure is fine, and that `configSecretName` is just an outlier that is probably misplaced

Basically, the "rationale" IMO is that `spec.service` is for flags that are passed to the operator. In this sense, `configSecretName` should've been placed directly under the `spec` key.

I've opened an issue to investigate whether that makes sense.

krnowak:
As I said in #269, I think that a rationale could be k8s specific stuff vs habitat specific stuff.

clintoncwolfe:
[...]

PR #184, opened 2018-02-07
Add Env key

asymmetric:
Adds an Env key to the Habitat.Spec type, which allows users to pass
environment variables to the container.

See #181.

### DIFF
--- a/examples/env-vars/README.md
+++ b/examples/env-vars/README.md
@@ -9,7 +9,9 @@ After the Habitat operator is up and running, execute the following command from
     kubectl create -f examples/env-vars/habitat.yml
 
 By default, Redis listens on port 6379, but we change this to 6999 by passing
-the `HAB_REDIS` environment variable to the supervisor.
+the `HAB_REDIS` environment variable to the supervisor (the supervisor
+automatically loads environment variables named `HAB_PACKAGENAME`, as explained
+[here](https://www.habitat.sh/docs/using-habitat/#config-updates)).
 
 When you inspect the logs of the pod that was just created (with `kubectl logs
 $podname`) you should see something like:
--- a/examples/env-vars/README.md
+++ b/examples/env-vars/README.md
@@ -1,0 +1,23 @@
+# Configuration via environment variables
+
+This example demonstrates how environment variables can be used for service configuration.
+
+## Workflow
+
+After the Habitat operator is up and running, execute the following command from the root of this repository:
+
+    kubectl create -f examples/env-vars/habitat.yml
+
+By default, Redis listens on port 6379, but we change this to 6999 by passing
+the `HAB_REDIS` environment variable to the supervisor.
+
+When you inspect the logs of the pod that was just created (with `kubectl logs
+$podname`) you should see something like:
+
+    redis.foobar(O): 42:M 07 Feb 15:34:08.765 * The server is now ready to accept connections on port 6999
+
+## Environment Variables
+
+You can use the full range of options defined by the Downward API. For an
+overview, please refer to [this
+guide](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/).
--- a/examples/env-vars/README.md
+++ b/examples/env-vars/README.md
@@ -1,6 +1,6 @@
 # Configuration via environment variables
 
-This example demonstrates how environment variables can be used for service configuration.
+This example demonstrates how environment variables can be used for initial service configuration.
 
 ## Workflow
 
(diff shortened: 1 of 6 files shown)


examples/envvars/README.md
```diff
@@ -0,0 +1,23 @@
+# Configuration via environment variables
+
+This example demonstrates how environment variables can be used for service configuration.
+
+## Workflow
+
+After the Habitat operator is up and running, execute the following command from the root of this repository:
+
+    kubectl create -f examples/config/habitat.yml
```
LiliC: s/config/envvars/

asymmetric:
I should probably add an E2E test.

asymmetric:
The test failure made me realize that we don't run the [`code-generator`](https://github.com/kubernetes/code-generator) stuff automatically, and we probably should.

I tried installing it with dep, but failed miserably:

    kubernetes-1.9.0: Could not introduce k8s.io/code-generator@kubernetes-1.9.0, as its subpackage k8s.io/code-generator does not contain usable Go code (*build.NoGoError).. (Package is required by (root).

 I'm not sure if it's a problem with dep, but I think we should run these scripts as part of the CI build.

asymmetric:
I seemed to have found out how to *temporarily* install this with dep:

    dep ensure -add k8s.io/code-generator/cmd/deepcopy-gen@kubernetes-1.9.0

In order to make it permanent, I'd have to add it as a dependency to our project with `import _`.

asymmetric:
But that of course doesn't build it. I now do `go get k8s.io/code-generator/cmd/deepcopy-gen`.

LiliC:
@asymmetric the command I used was: 
```
./generate-groups.sh deepcopy github.com/kinvolk/habitat-operator/pkg/client github.com/kinvolk/habitat-operator/pkg/apis "habitat:v1beta1""
```

Tried it locally and it works for me.

LiliC:
What I have seen other projects do, is they add the script to the `hacks` directory for example, and run it from there next time they need it.

examples/envvars/README.md
```diff
@@ -0,0 +1,23 @@
...
+the `HAB_REDIS` environment variable to the supervisor.
+
+When you inspect the logs of the pod that was just created (with `kubectl logs
+$podname`) you should see something like:
+
+    redis.foobar(O): 42:M 07 Feb 15:34:08.765 * The server is now ready to accept connections on port 6999
+
+## Environment Variables
+
+You can use the full range of options defined by the Downward API. For an
```
LiliC: As per the agreement, please don't break the lines in the file manually. :)

examples/envvars/habitat.yml
```diff
@@ -0,0 +1,16 @@
+apiVersion: habitat.sh/v1beta1
+kind: Habitat
+metadata:
+  name: example-envvars-habitat
+spec:
+  # the core/redis habitat service packaged as a Docker image
+  image: kinvolk/redis-hab
+  count: 1
+  env:
```
LiliC: Can this be under the `service`, or does it have to be under `spec`?

examples/envvars/README.md
asymmetric: It makes navigation in vim hard when lines are super long. Is there a problem with me breaking lines?

examples/envvars/habitat.yml
asymmetric: That's a [good question](https://github.com/kinvolk/habitat-operator/issues/182).

examples/envvars/README.md
asymmetric: They still show up as one line on the rendered Markdown. It's just for my convenience. Or does it inconvenience people using other editors?

zeenix: > It makes navigation in vim hard when lines are super long

Do you know of 'g' prefix to nav keys? As in 'gj' instead of 'j'.

asymmetric: Yes, I do :)

examples/envvars/habitat.yml
asymmetric: I'm open to suggestions on where we should place the key. What's the benefit/rationale of placing it under service? Feel free to add to the linked discussion for more general points.

LiliC: Well the `service` is where we place things that are specific to this service. But then again we also have the `image` and `count` not under there. So no clue. Up to you.

asymmetric: See https://github.com/kinvolk/habitat-operator/issues/182#issuecomment-365275730. I think we can make the change in another PR.

asymmetric:
Is this good to go? Or is the linebreaks thing an actual issue? :)

examples/envvars/README.md
```diff
@@ -0,0 +1,23 @@
+# Configuration via environment variables
+
+This example demonstrates how environment variables can be used for service configuration.
```
LiliC: Think this is for initial configuration only, correct?

examples/envvars/README.md
LiliC: Think you forgot to update this? Or am I missing something

examples/env-vars/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+This example demonstrates how environment variables can be used for initial service configuration.
+
+## Workflow
+
+After the Habitat operator is up and running, execute the following command from the root of this repository:
+
+    kubectl create -f examples/env-vars/habitat.yml
+
+By default, Redis listens on port 6379, but we change this to 6999 by passing
```
LiliC: Think you can mention something about the `HAB_SVC_NAME` and point to the Habitat docs here? As `HAB_REDIS` doesn't tell me that is the actual service/pkg name.

PR #187, opened 2018-02-14
Improvements to Helm README

asymmetric:
* Removes TL;DR
* Removes instructions for v1.5.x

See commit messages for why.

### DIFF
--- a/helm/habitat-operator/README.md
+++ b/helm/habitat-operator/README.md
@@ -2,13 +2,6 @@
 
 Installs [habitat-operator](https://github.com/kinvolk/habitat-operator) to manage Habitat services in a Kubernetes cluster.
 
-## TL;DR;
-
-```console
-$ helm repo add habitat https://kinvolk.github.io/habitat-operator/helm/charts/stable/
-$ helm install habitat/habitat-operator
-```
-
 ## Introduction
 
 This chart bootstraps a [habitat-operator](https://github.com/kinvolk/habitat-operator) deployment in a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
@@ -42,6 +35,7 @@ $ helm init --service-account tiller
 To install the chart with the release name `my-release`:
 
 ```console
+$ helm repo add habitat https://kinvolk.github.io/habitat-operator/helm/charts/stable/
 $ helm install --name my-release habitat/habitat-operator
 ```
 
--- a/helm/habitat-operator/README.md
+++ b/helm/habitat-operator/README.md
@@ -8,6 +8,9 @@ This chart bootstraps a [habitat-operator](https://github.com/kinvolk/habitat-op
 
 ## Prerequisites
 
+Note: If you have Kubernetes and Helm installed, skip to [this
+section](#Installing the Chart).
+
 See [Habitat Operator's README](https://github.com/kinvolk/habitat-operator/blob/master/README.md).
 
 ### RBAC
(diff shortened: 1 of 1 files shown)


helm/habitat-operator/README.md
```diff
@@ -2,13 +2,6 @@
 
 Installs [habitat-operator](https://github.com/kinvolk/habitat-operator) to manage Habitat services in a Kubernetes cluster.
 
-## TL;DR;
-
-```console
-$ helm repo add habitat https://kinvolk.github.io/habitat-operator/helm/charts/stable/
-$ helm install habitat/habitat-operator
-```
-
```
zeenix: You also need a working k8s cluster but we won't tell how to set that up here. It's nice to have the helm setup instructions here for folks who haven't yet set it up before they use this particular chart, i don't think we can assume that's a majority case. So it's good to have a shortcut for folks who already have helm setup and they just now want to try out habitat through helm.

Instead let's add a small comment before these commmands. Something like "If you already have helm setup in your kubernetes cluster, you can simply use the following to install the Habitat operator'.

asymmetric: We don't support 1.5 clusters though, so people cannot run the operator there.

helm/habitat-operator/README.md
```diff
@@ -32,6 +32,7 @@ $ kubectl create clusterrolebinding tiller --clusterrole cluster-admin --service
 ```
 
 3. Install Tiller, specifying the new ServiceAccount
+
```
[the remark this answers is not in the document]
asymmetric: I assume you're trolling :trollface:.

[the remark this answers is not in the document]
zeenix: nope, bikeschedding.

helm/habitat-operator/README.md
asymmetric: You're right, I removed this because I thought the same info was repeated below. My only issue is with the order, and with the TL;DR title, because it's misleading (see commit message).

i'll add the info back below.

zeenix: Good that we agree but the placement is correct in here. tl'dr like this is always in the beginning and the point is to allow people to not have to read more than they need to.

asymmetric: I don't think it's actually a TL;DR, for the reasons explained in the commit message.

zeenix: It very much is tl'dr. People who have helm setup, should just be able to skip rest of the README. But if you feel strongly also, at least add a "if you have helm setup, please skip to LINK" in this point.

asymmetric: I went through this yesterday. I was in a hurry, I saw a TL;DR and I assumed I could skip the rest. Wasn't the case.

I can add a "skip to link" thing.

asymmetric:
PTALA.

krnowak:
LFAD.

helm/habitat-operator/README.md
```diff
@@ -8,6 +8,9 @@ This chart bootstraps a [habitat-operator](https://github.com/kinvolk/habitat-op
 
 ## Prerequisites
 
+Note: If you have Kubernetes and Helm installed, skip to [this
+section](#Installing the Chart).
```
zeenix: Let's be consistent with whether or not to break long lines.

zeenix:
apart from the line breaking nit, it looks good to me.

asymmetric:
Line unbroken.

PR #188, opened 2018-02-14
Use regex to find all the YAML files

zeenix:
The rule to update the version string all the YAML files was broken from
4d7775336c4f9da28c752809797f6a42d7bd0942, where we tried to use the '-o'
option of `find` command replace version string in both '*.yaml' and
'*.yml' files. Turns out, that doesn't work for some reason.

So let's just use a regular expression with `find` to keep things simpler
and make the rule work again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,13 +28,16 @@ clean-test:
 	kubectl delete clusterrole habitat-operator
 
 update-version:
-	find examples helm -name "*.yml" -o -name "*.yaml" -type f \
+	find examples -name "*.yml" -type f \
 		-exec sed -i.bak \
 		-e "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" \
+		'{}' \;
+	find helm -name "*.yaml" -type f \
+		-exec sed -i.bak \
 		-e "s/tag:.*/tag: v$$(cat VERSION)/g" \
 		-e "s/version:.*/version: $$(cat VERSION)/g" \
 		'{}' \;
-	find examples helm -name "*.yml.bak" -o -name "*.yaml.bak" -type f \
+	find examples helm -regex '.*\.ya?ml$$.bak' -type f \
 		-exec rm '{}' \;
 
 .PHONY: build test linux image e2e clean-test update-version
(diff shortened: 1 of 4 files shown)


examples/crd.yml
```diff
@@ -7,7 +7,7 @@ spec:
   # group name to use for REST API: /apis/<group>/<version>
   group: habitat.sh
   # version name to use for REST API: /apis/<group>/<version>
-  version: v1
+  version: 0.5.0
```
asymmetric: The CRD should actually be `v1beta1`, we must've forgotten to update it.

Makefile
```diff
@@ -28,13 +28,16 @@ clean-test:
 	kubectl delete clusterrole habitat-operator
 
 update-version:
-	find examples helm -name "*.yml" -o -name "*.yaml" -type f \
+	find examples -name "*.yml" -type f \
 		-exec sed -i.bak \
 		-e "s/habitat-operator:.*/habitat-operator:v$$(cat VERSION)/g" \
+		'{}' \;
+	find helm -name "*.yaml" -type f \
```
asymmetric: Wut? This doesn't make a lot of sense. Why execute both commands twice instead of using the regex?

Makefile
```diff
@@ -28,13 +28,16 @@ clean-test:
...
+		-exec sed -i.bak \
 		-e "s/tag:.*/tag: v$$(cat VERSION)/g" \
 		-e "s/version:.*/version: $$(cat VERSION)/g" \
 		'{}' \;
-	find examples helm -name "*.yml.bak" -o -name "*.yaml.bak" -type f \
+	find examples helm -regex '.*\.ya?ml$$.bak' -type f \
```
asymmetric: Are you mixing approaches?

Makefile
zeenix: because of the "version: " expression. Recall that it modifies the API version in the crd example.

Makefile
zeenix: yeah cause there is nothing wrong with the regex itself.

PR #191, opened 2018-02-15
Simplify gh-pages branch

krnowak:
This drops all the unnecessary stuff (code, examples and whatnot), moves the README.md and the script from `helm/charts` to toplevel directory and updates the script to perform the chart file generation.


zeenix:
but your changes will change the URL at which charts will be served and we don't want that. Let's please keep the hierarchy as is.

krnowak:
I didn't change the hierarchy at all.

zeenix:
Ah you only moved the script and README. That's fine I guess. We'll need these changes merged to properly test them i guess.

PR #192, opened 2018-02-15
Fix a path to gh-pages README.md

krnowak:
Is was moved to the toplevel directory as a part of gh-pages branch
cleanup.

To be merged after #191.

### DIFF
--- a/release.md
+++ b/release.md
@@ -29,7 +29,7 @@ In the root directory of the repository generate the Docker image and push it to
 
 ## Generate the Helm chart
 
-Switch to `gh-pages` branch and follow the steps in the `helm/charts/README.md` file.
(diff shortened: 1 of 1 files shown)


release.md
```diff
@@ -29,7 +29,7 @@ In the root directory of the repository generate the Docker image and push it to
 
 ## Generate the Helm chart
 
-Switch to `gh-pages` branch and follow the steps in the `helm/charts/README.md` file.
+Switch to `gh-pages` branch and follow the steps in the toplevel `README.md` file.
```
LiliC: Don't think the toplevel part is needed?

krnowak: Sure, can drop it.

PR #194, opened 2018-02-15
Fix apiVersion in deployment manifest template

zeenix:
It should be the same as the apiVersion in examples/habitat-operator-deployment.yml

### DIFF
--- a/helm/habitat-operator/templates/deployment.yaml
+++ b/helm/habitat-operator/templates/deployment.yaml
@@ -1,4 +1,4 @@
-apiVersion: apps/v1beta1
+apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
   labels:


zeenix:
Maybe it's not needed per say and I confused things but shouldn't we keep the apiVersion both deployment files the same?

LiliC:
@zeenix if you insist on it, sure we can use this, as I said it's the same. 

> shouldn't we keep the apiVersion both deployment files the same?

They are two separate things, if you use Helm to deploy the operator you won't need the other deployment file anyways. So not sure I understand your concern here?

zeenix:
>     shouldn't we keep the apiVersion both deployment files the same?
> 
> They are two separate things, if you use Helm to deploy the operator you won't need the other deployment file anyways. So not sure I understand your concern here?

Not for any functional benefits but consistency and avoiding any confusions.

asymmetric:
What should we do with this one?

asymmetric:
I have no clue to be honest. Up to you.

PR #199, opened 2018-03-01
Some corrections in API docs

krnowak:
Name field was missing in Service. Env field is a part of Spec, not Service.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -17,6 +17,7 @@ The following is a description of the Habitat operator API. To see manifest exam
 | count | Count is the amount of Services that should start in Habitat. | int | true |
 | image | Image is the Docker image of the Habitat Service. | string | true |
 | service |  | [Service](#service) | true |
+| env | A list of environment variables | [][EnvVar](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#envvar-v1-core) | false
 
 ## Service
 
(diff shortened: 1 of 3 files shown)


docs/api.md
```diff
@@ -27,7 +28,7 @@ The following is a description of the Habitat operator API. To see manifest exam
 | configSecretName | configSecretName is the name of the Kubernetes Secret containing the config file - user.toml - that the user has previously created. Habitat will use it for initial configuration of the service. | string | false |
 | ringSecretName | The name of the Kubernetes Secret that contains the ring key, which encrypts the communication between Habitat supervisors. | string | false |
 | bind | When one service connects to another forming a producer/consumer relationship. Able to specify multiple binds. | [][Bind](#bind) | false |
-| env | A list of environment variables | [][EnvVar](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#envvar-v1-core) | false
+| name | A name part of the Habitat package identification (so it is "nginx" for "core/nginx/1.11.10/20171014214153") | string | true |
```
LiliC: Maybe just add the following as to why this is a requirement:
```
This field is used to mount the `user.toml` file in the correct directory under /hab/svc/ in the Pod.
```

Feel free to rephrase the above, it's copy pasta from the https://github.com/kinvolk/habitat-operator/blob/master/pkg/apis/habitat/v1beta1/types.go#L80

krnowak: And that comment is also wrong about the `/hab/svc` path too. ;) Will amend.

Issue #200, opened 2018-03-06
Use Status field of Habitat object


asymmetric:
The `Habitat`'s field should reflect the state of the underlying resources.

For example, we might decide that it should have a status of `"Creating"` until all `Pod`s are in a `"Running"` state.

This would make the overall system easier to observe, as by just looking at a `"Habitat"`s state users would be able to understand what's going on.

asymmetric:
Couple of points:

The `status` field of `StatefulSet`s is not used to describe the status of the whole set of sub-resources. i.e. it does not say if e.g. the underlying `PVC`s are in a `Pending` state. This is the original usecase I had thought of, and the fact it's not done by upstream is interesting. Need to investigate more.

What upstream *does* do is populate `Event`s on the `STS` that mention its sub-resources. This is an orthogonal feature which I think we can implement right away.

krnowak:
Normally when you call `kubectl get pods` you get something like

```
NAME                           READY     STATUS    RESTARTS   AGE
example-standalone-habitat-0   1/1       Running   0          18h
```

For `kubectl get habitats` we get:
```
NAME                         AGE
example-standalone-habitat   18h
```

Would be nice to get the `STATUS` column there too.

You can also use `kubectl describe pod example-standalone-habitat-0`, it should give you something like:

```
Name:           example-standalone-habitat-0
Namespace:      default
Node:           kube-spawn-default-worker-k8q4ug/10.22.0.28
Start Time:     Thu, 03 May 2018 16:57:22 +0200
Labels:         controller-revision-hash=example-standalone-habitat-777f4f9c9c
                habitat=true
                habitat-name=example-standalone-habitat
                statefulset.kubernetes.io/pod-name=example-standalone-habitat-0
                topology=standalone
Annotations:    <none>
Status:         Running
IP:             10.32.0.3
Controlled By:  StatefulSet/example-standalone-habitat
…
```

`kubectl describe hab example-standalone-habitat` doesn't have the `Status:` line for now.

The status stuff:

- [as a member of habitat struct](https://github.com/habitat-sh/habitat-operator/blob/65fdbb8235d392197614adf1149dacc39385ae8d/pkg/apis/habitat/v1beta1/types.go#L43)
- [a type itself](https://github.com/habitat-sh/habitat-operator/blob/65fdbb8235d392197614adf1149dacc39385ae8d/pkg/apis/habitat/v1beta1/types.go#L95-L100)

Some reading:
[...]

PR #201, opened 2018-03-06
Implement persistence with StatefulSets

### LINKED ISSUE #38: Use ownerReferences with CRD
Using `OwnerReferences` allows us to define relationships between Resources, so that deleting an owner can automatically delete owned resources.

Currently, we use `OwnerReferences` to associate a ConfigMap with a Deployment.

It could be useful to make all Resources we create dependent on the CustomResource, but there might currently be [problems](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#known-issues) with that.

### LINKED ISSUE #44: Use StatefulSets instead of Deployment
Currently we are using [Deployments](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) to deploy our Habitat Service, but since we do not know what are deploying, what type of service that is, it could be anything from a DB to a simple Rails application. We should not just assume our Habitat service would be stateless.

Couple of advantages of [StatefulSets](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/):
- Graceful deployment and scaling
- Stable network identity
- Graceful deletion and termination
- Stable, persistent storage

These would be very useful especially if our service is for example a DB.

### LINKED ISSUE #181: RFE: support env and volumeMounts in HabitatSpec

asymmetric:
This PR

* Changes the type of object we create from `Deployment` to `StatefulSet`
* Adds a `Persistence` key to the `Habitat` type
* Starts splitting `controller.go` into separate modules
* Removes `docs/api.md` and just points users to `types.go`
* ~Adds tests for persistence~

The main issue with using `Deployment`s for persistence was that there was no automatic way to assign a different PVC to each Pod, so we ended up with Pods that shared the same Volume.

See #44 and #195 for more in-depth discussions.

**NOTE**: Tests are included but skipped, as they don't work in our current testing environment (it's not possible to provision `PersistentVolume`s when running minikube with `--vm-driver=none`).

Closes #44.
Closes #38.
Closes #181.


asymmetric:
@LiliC could you run the deepcopy script against this branch, and see if there's any change?

jeremymv2:
sosweet

examples/persisted/habitat.yml
```diff
@@ -0,0 +1,25 @@
...
+    # the size of the volume that will be mounted in each Pod
+    size: 1Gi
+    # a StorageClass object with name "standard" must be created beforehand by
+    # the cluster administrator
+    storageClassName: standard
+    # the location under which the volume will be mounted
+    mountPath: /tmp/foobar
+  env:
+    - name: HAB_REDIS
+      value: '{ "protected-mode": "no" }'
```
LiliC: Please add a note of warning here.

pkg/apis/habitat/v1beta1/types.go
```diff
@@ -51,6 +51,18 @@ type HabitatSpec struct {
 	// The EnvVar type is documented at https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.9/#envvar-v1-core.
 	// Optional.
 	Env []corev1.EnvVar `json:"env,omitempty"`
+	// Optional.
+	Persistence *Persistence `json:"persistence,omitempty"`
+}
+
+type Persistence struct {
```
LiliC: Can we call this `Storage` instead? Things make a bit more sense, for example you then have `Size`, `ClassName` as part of `Storage`. WDYT?

pkg/controller/controller.go
```diff
@@ -922,11 +679,20 @@ func habitatKeyFromLabeledResource(r metav1.Object) (string, error) {
 func newConfigMap(ip string, h *habv1beta1.Habitat) *apiv1.ConfigMap {
 	return &apiv1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
+			// TODO the name should also change to support multiple habitats
```
LiliC: Can you explain this `TODO` a bit more?

pkg/controller/stateful_sets.go
```diff
@@ -0,0 +1,325 @@
...
+	d, ok := obj.(*appsv1beta1.StatefulSet)
+	if !ok {
+		level.Error(hc.logger).Log("msg", "Failed to type assert StatefulSet", "obj", obj)
+		return
+	}
+
+	h, err := hc.getHabitatFromLabeledResource(d)
+	if err != nil {
+		// Could not find Habitat, it must have already been removed.
+		level.Debug(hc.logger).Log("msg", "Could not find Habitat for StatefulSet", "name", d.Name)
```
LiliC: Hmm this msg doesn't make much sense to me as a consumer, maybe adding `Could not find Habitat resource for StatefulSet` would be better.

test/e2e/operator_test.go
```diff
@@ -183,3 +187,67 @@ func TestHabitatDelete(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestHabitatPersistence(t *testing.T) {
+	// We run minikube in a VM on Travis. In that environment, we cannot create PersistentVolumes.
```
LiliC: Can you explain this a bit, what is the reason behind this?

Also if it doesn't work why do we have this test? I am a bit confused.

examples/rbac/rbac.yml
```diff
@@ -34,7 +34,7 @@ rules:
 - apiGroups:
   - apps
   resources:
-  - deployments
```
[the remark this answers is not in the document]
asymmetric: No I don't think so, as it's not the operator that's creating `Volume`s, it's the `StatefulSet`.

pkg/apis/habitat/v1beta1/zz_generated.deepcopy.go
```diff
@@ -142,6 +151,22 @@ func (in *HabitatStatus) DeepCopy() *HabitatStatus {
 	return out
 }
 
+// DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
```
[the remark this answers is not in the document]
asymmetric: I didn't :) This is your commit actually, from the `Deployment`s PR.

pkg/apis/habitat/v1beta1/types.go
asymmetric: Not sure that `Storage` by itself makes it clear that this is persistent rather than volatile.

pkg/controller/controller.go
asymmetric: Oops, that was just for me. Anyway, I think that making the CM name a constant has the drawback that we can only have one. In the future, this could pose problems if we have more than one ring in the same namespace.

pkg/controller/stateful_sets.go
asymmetric: Well, it just uses the same format as every other object. We don't say "StatefulSet resource" after all.

test/e2e/operator_test.go
asymmetric: The reason is that since we run the operator in containers on Travis, the dynamic volume provisioning that Minikube normally does (when it's on a VM) doesn't work.

I think leaving the test in makes sense because:

* it took time to write it, so we would lose that by deleting it
* we can reactivate it when we move away from testing with minikube (which I think we should)
* it documents that there's a problem and what it is

pkg/controller/stateful_sets.go
LiliC: Yes understand, but it can be confusing for a user as the name conveys two things. :) Anyways, more of a nit for clarity.

pkg/controller/controller.go
LiliC: The problem with is that is how would we handle binding if they are not in one ring. That's why we decided to go with one ring...

test/e2e/operator_test.go
LiliC: Personally don't like dead code in master, maybe you could just cherry-pick this and keep it in a branch so the work is not lost.

asymmetric: I think that if it's marked as `skip`, then it's fine. It's common for tests to be temporarily marked as skippable IMO. It would be different if it was business logic, but I think it's fine for tests. I'd rather keep this in unless it's blocking for you.

pkg/apis/habitat/v1beta1/types.go
LiliC: `PersistantStorage`? :)

asymmetric: I'm afraid you're gonna have to re-run the script though :)

pkg/controller/controller.go
asymmetric: Yes, I didn't mean to say that I thought it through, that's why it was just there for me as a reminder.

LiliC:
@asymmetric I wanted to try this out but can't seem to compile it locally. But it seems to be running fine on travis. :/ Does `make build` work for you?

pkg/controller/persistent_volumes.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	_, exists, err := hc.habInformer.GetStore().GetByKey(key)
+	if err != nil {
+		level.Error(hc.logger).Log("msg", "Failed to get key in Store", "obj", key)
+		return
+	} else if !exists {
+		level.Debug(hc.logger).Log("msg", "No matching Habitat found for PVC", "name", pv.Name)
+		return
+	} else {
```
LiliC: I think we can get rid of this else as we return in other cases anyways.

pkg/controller/persistent_volume_claims.go
```diff
@@ -0,0 +1,88 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
```
LiliC: It's 2018 :)

pkg/controller/persistent_volumes.go
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
```
LiliC: It's 2018 :)

pkg/controller/stateful_sets.go
```diff
@@ -0,0 +1,325 @@
+// Copyright (c) 2017 Chef Software Inc. and/or applicable contributors
```
LiliC: It's 2018 :)

pkg/controller/persistent_volumes.go
asymmetric: I think in this case it makes more sense to keep it in in this case, because it's a 3-way situation we're dealing with, and having them together helps (me) map the idea to the code better. 

Usually with 2-branch conditionals that's not necessary, but I find that it helps here.

WDYT?

pkg/controller/persistent_volumes.go
```diff
@@ -0,0 +1,69 @@
...
+
+	hc.pvInformer = cache.NewSharedIndexInformer(
+		source,
+		&apiv1.PersistentVolume{},
+		resyncPeriod,
+		cache.Indexers{},
+	)
+
+	hc.pvInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: hc.handlePVDelete,
```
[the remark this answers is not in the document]
asymmetric: I think that at least for now, we can avoid implementing resizes of the storage: it's an optional feature that introduces quite a lot of [complexity](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#expanding-persistent-volumes-claims). I'd wait for this feature to have stabilized a bit (and for someone to request it) before we figure out how to do it best.

But we should definitely make this clear and notify the user somehow, when they change the `Size` in the manifest, that it's not supported.

asymmetric:
About the `docs/api.md` file: is it useful? I keep forgetting to keep it in sync, and I'd like to know if there's a good reason to do that.

LiliC:
> About the docs/api.md file: is it useful? I keep forgetting to keep it in sync, and I'd like to know if there's a good reason to do that.

I just find the documentation helpful when needing to reference something, but we could also implement a helper script for that? Or just get rid of it, up to you. :)

pkg/controller/persistent_volumes.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/go-kit/kit/log/level"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+func (hc *HabitatController) cachePersistentVolumes() {
+	source := newListWatchFromClientWithLabels(
+		hc.config.KubernetesClientset.CoreV1().RESTClient(),
+		"persistentvolumes",
+		apiv1.NamespaceAll,
```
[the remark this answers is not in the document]
asymmetric: I guess so, because users can create `Habitat`s in any namespace. Why would this case be different from other resources we watch?

pkg/controller/persistent_volumes.go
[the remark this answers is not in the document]
LiliC: Then IMO I don't see the value of caching PV informer or rather reacting to the changes. Same for the PVC informer. Clean up of volumes is intentionally manual and the rest is already handled by the StatefulSet. The user is already informed, via monitoring, logs, etc., so I feel like we are just duplicating that here. Especially since we are caching ALL the PVC/PV, that can be a huge number in a normal cluster.

pkg/controller/controller.go
```diff
@@ -927,6 +684,14 @@ func newConfigMap(ip string, h *habv1beta1.Habitat) *apiv1.ConfigMap {
 			Labels: map[string]string{
 				habv1beta1.HabitatLabel: "true",
 			},
+			OwnerReferences: []metav1.OwnerReference{
+				metav1.OwnerReference{
+					APIVersion: fmt.Sprintf("%s/%s", habv1beta1.GroupName, habv1beta1.Version),
```
LiliC: There is `h.APIVersion` you can use.

pkg/controller/controller.go
```diff
@@ -927,6 +684,14 @@ func newConfigMap(ip string, h *habv1beta1.Habitat) *apiv1.ConfigMap {
...
+					Kind:       "Habitat",
```
LiliC: Same here there is `h.Kind` you can use.

pkg/controller/persistent_volumes.go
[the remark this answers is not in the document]
asymmetric: We are only caching ones with label `habitat: true`. Also, the point of caching is that they could be  removed by anything/anybody else, and we need to be able to notify the user (and in the future, update the `Status`).

According to the Controller Bible, verse 7:

> There are other actors in the system. Just because you haven't changed an object doesn't mean that somebody else hasn't.

pkg/controller/stateful_sets.go
```diff
@@ -0,0 +1,325 @@
...
+		// Pass --bind flag.
+		bindArg := fmt.Sprintf("%s:%s.%s", bind.Name, bind.Service, bind.Group)
+		habArgs = append(habArgs,
+			"--bind", bindArg)
+	}
+
+	base := &appsv1beta1.StatefulSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: h.Name,
+		},
```
LiliC: Add a `OwnerReference` to the `StatefulSet` as well, like above in the `ConfigMap`.

pkg/controller/controller.go
```diff
@@ -927,6 +684,14 @@ func newConfigMap(ip string, h *habv1beta1.Habitat) *apiv1.ConfigMap {
...
+					Name:       h.Name,
+					UID:        h.UID,
+				},
```
LiliC: Also I noticed we do not say anything about `BlockOwnerDeletion`, I would add this here. It is very useful IMO.

pkg/controller/stateful_sets.go
asymmetric: We currently do that [here](https://github.com/habitat-sh/habitat-operator/blob/aaca35c5dee80138e793f465ebbd1ddc95d2b10f/pkg/controller/controller.go#L485), but I agree we could simply delegate this to Kubernetes.

pkg/controller/controller.go
asymmetric: > if the owner has the "foregroundDeletion" finalizer

Is this our case?

It's not entirely clear for me if any defaults are defined for the owning resource (`Habitat`). Otherwise, it seems to me that `propagationPolicy` could be specified by a user in their `DELETE` request. Is this something we should be worried about?

pkg/controller/persistent_volumes.go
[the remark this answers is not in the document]
LiliC: > and we need to be able to notify the user

But that is already done by its respective controller.

Agree that we should react, but we are not doing anything apart from logging. So I don't see the point of it, IMHO. :)

pkg/controller/stateful_sets.go
asymmetric: Done.

pkg/controller/controller.go
LiliC: Not sure if it's something to worry about, just thought it was nice to have.

pkg/controller/persistent_volumes.go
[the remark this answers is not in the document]
asymmetric: We will update the `Status` in #200 :)

pkg/controller/persistent_volumes.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// Find out if there's a Habitat object around
+	// If so, we have a problem.
+	key, err := cache.MetaNamespaceKeyFunc(pv.Spec.ClaimRef)
+	if err != nil {
+		level.Error(hc.logger).Log("msg", "Failed to get key", "obj", pv)
+		return
+	}
+
+	_, exists, err := hc.habInformer.GetStore().GetByKey(key)
```
krnowak: That part of code looks like a copy pasta of the `findHabForPVC` function. Maybe just call it here?

pkg/controller/stateful_sets.go
```diff
@@ -0,0 +1,332 @@
...
+
+		secretVolumeMount := &apiv1.VolumeMount{
+			Name: userConfigFilename,
+			// The Habitat supervisor creates a directory for each service under /hab/svc/<servicename>.
+			// We need to place the user.toml file in there in order for it to be detected.
+			MountPath: fmt.Sprintf("/hab/user/%s/config", h.Spec.Service.Name),
+			ReadOnly:  false,
+		}
+
+		base.Spec.Template.Spec.Containers[0].VolumeMounts = append(base.Spec.Template.Spec.Containers[0].VolumeMounts, *secretVolumeMount)
```
krnowak: You could probably do `spec := &base.Spec.Template.Spec` and then `spec.Containers[0].VolumeMounts = …` to drop a bit of a repeating the same thing.

krnowak: Or even move this `spec := …` outside this block, so it could be also used for mounting the persistent volume or the ring secret volume.

pkg/controller/controller.go
```diff
@@ -71,14 +70,18 @@ type HabitatController struct {
...
+	pvcInformer cache.SharedIndexInformer
 
 	// cache.InformerSynced returns true if the store has been synced at least once.
-	habInformerSynced    cache.InformerSynced
-	deployInformerSynced cache.InformerSynced
-	cmInformerSynced     cache.InformerSynced
+	habInformerSynced cache.InformerSynced
+	stsInformerSynced cache.InformerSynced
+	cmInformerSynced  cache.InformerSynced
+	pvInformerSynced  cache.InformerSynced
```
krnowak: What are `pvInformerSynced` and `pvcInformerSynced` for? They seem to be unused. Did you forget to use them in `WaitForCacheSync` call?

pkg/controller/persistent_volumes.go
asymmetric: Actually, I think there are 2 problems with my approach here:

* We are not telling users that they `PersistentVolume`s they create (if they're going with static provisioning) need to have the `Habitat` label
* We find the Habitat by name, which seems like a weak approach. Maybe a label would be better? Considering that the PV might be created beforehand, it seems weird to force the user to know already what the name should be. A label, OTOH, can be added later, and seems a better approach than asking the user to rename the PV.

pkg/controller/persistent_volumes.go
asymmetric: I ended up agreeing with you :)

examples/rbac/rbac.yml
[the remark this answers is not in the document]
asymmetric: You were right btw, we did have to: the ListWatcher running in the operator means we need to assign `list` and `watch` permissions.

asymmetric:
PTALA, I reverted all commits that dealt with watching PVs/PVCs. I haven't removed them from history because I thought it would be nice to have them for when we'll re-implement this.

Let me know what you think.

pkg/controller/controller.go
```diff
@@ -347,15 +274,15 @@ func (hc *HabitatController) enqueueCM(obj interface{}) {
 }
 
 func (hc *HabitatController) handleCMAdd(obj interface{}) {
-	hc.enqueueCM(obj)
+	hc.handleCM(obj)
```
LiliC: Not sure if this is in the scope of this PR or not, but since you are redoing the logic of resources being created or not, I open an issue about CM not being deleted when Habitat resource is deleted, so maybe it's something worth looking into. It can be outside of this PR as well, up to you...

pkg/controller/stateful_sets.go
```diff
@@ -0,0 +1,335 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package controller
+
+import (
+	"fmt"
+
+	"github.com/go-kit/kit/log/level"
+	habv1beta1 "github.com/habitat-sh/habitat-operator/pkg/apis/habitat/v1beta1"
```
LiliC: My personal preference is to have the these imports separated with a line break from the outside imports.

pkg/controller/stateful_sets.go
```diff
@@ -0,0 +1,335 @@
...
+	appsv1beta1 "k8s.io/api/apps/v1beta1"
+	apiv1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+const persistentVolumeName = "persistent"
```
LiliC: Why is this name hardcoded?

test/e2e/framework/helpers.go
```diff
@@ -223,7 +224,7 @@ func ConvertSecret(pathToYaml string) (*v1.Secret, error) {
 	return &s, nil
 }
 
-func convertToK8sResource(pathToYaml string, into interface{}) error {
+func convertToK8sResource(pathToYaml string, into runtime.Object) error {
```
LiliC: cool! TIL

LiliC:
BTW since this is a huge breaking change IMO we really need to bump up the version here of the API, so to `v2beta1` or something like that. As that we don't get problems if this operator is deployed into a cluster where Habitat resources, e.g. Deployments already exist as that cannot be handled by this operator.

examples/rbac/rbac.yml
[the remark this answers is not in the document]
asymmetric: Exactly.

pkg/controller/stateful_sets.go
asymmetric: Because it never changes ;)

It's a bit like the name for the CM volume. We hardcode it, and since we are good programmers, we move it to a constant ;)

Basically, there's no reason to expose this name and make it configurable I think.

pkg/controller/stateful_sets.go
asymmetric: Because it never changes ;)

It's a bit like the name for the CM volume. We hardcode it, and since we are good programmers, we move it to a constant ;)

Basically, there's no reason to expose this name and make it configurable I think.

pkg/controller/stateful_sets.go
asymmetric: Your wish is my command.

pkg/controller/controller.go
asymmetric: :+1:. I'll tackle it in another PR.

asymmetric:
We agreed offline to keep the current `Deployment`-based behavior in `v1beta1`, alongside the new `StatetfulSet`-based one on `v1beta2`. This will be done in a folloging PR.

So please go ahead and do your final reviews :)

pkg/apis/habitat/v1beta1/types.go
```diff
@@ -51,6 +51,20 @@ type HabitatSpec struct {
...
+// PersistentStorage contains the details of the persistent storage that the
+// cluster should provision.
+type PersistentStorage struct {
+	// Size is the volume's size.
+	// It uses the same format as Kubernetes' size fields, e.g. 10Gi
+	Size string `json:"size"`
+	// MountPath is the path at which the PersistentVolume will be mounted.
+	MountPath string `json:"mountPath"`
+	// StorageClassName is the name of the StorageClass that the StatefulSet will request.
+	StorageClassName string `json:"storageClassName"`
```
LiliC: Nit: Can we name this just `ClassName` as it's under the `PersistantStorage` anyways. WDYT?

pkg/apis/habitat/v1beta1/types.go
```diff
@@ -51,6 +51,20 @@ type HabitatSpec struct {
...
+	// Optional.
+	PersistentStorage *PersistentStorage `json:"persistentStorage,omitempty"`
+}
+
+// PersistentStorage contains the details of the persistent storage that the
+// cluster should provision.
+type PersistentStorage struct {
+	// Size is the volume's size.
+	// It uses the same format as Kubernetes' size fields, e.g. 10Gi
+	Size string `json:"size"`
```
LiliC: You are missing the `omitempty`, or are these are required? Then we should document that as we don't have the api docs.md anymore. :)

pkg/controller/controller.go
```diff
@@ -927,6 +652,14 @@ func newConfigMap(ip string, h *habv1beta1.Habitat) *apiv1.ConfigMap {
 			Labels: map[string]string{
 				habv1beta1.HabitatLabel: "true",
 			},
+			OwnerReferences: []metav1.OwnerReference{
+				metav1.OwnerReference{
+					APIVersion: habv1beta1.SchemeGroupVersion.String(),
+					Kind:       habv1beta1.HabitatKind,
```
LiliC: What do we use this OwnerReference for now?

examples/persisted/README.md
```diff
@@ -0,0 +1,37 @@
+# Persistent Storage example
```
LiliC: So the whole point of persistence is that you can mount already existing volumes, can you maybe write up some steps of how a user can do that.

examples/persisted/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Habitat objects are translated by the operator into `StatefulSet` objects, which
```
LiliC: Nit: Maybe add a note that the user needs to know about basic understanding of persistent volumes in K8s and link to it?

pkg/controller/stateful_sets.go
```diff
@@ -0,0 +1,336 @@
...
+
+func (hc *HabitatController) newStatefulSet(h *habv1beta1.Habitat) (*appsv1beta1.StatefulSet, error) {
```
LiliC: You know now that you extracted this you can actually write a unit test for this. WDYT?

pkg/controller/controller.go
```diff
@@ -802,15 +526,16 @@ func (hc *HabitatController) processNextItem() bool {
 
 // conform is where the reconciliation takes place.
 // It is invoked when any of the following resources get created, updated or deleted:
-// Habitat, Pod, Deployment, ConfigMap.
+// Habitat, Pod, StatefulSet, ConfigMap.
 func (hc *HabitatController) conform(key string) error {
 	obj, exists, err := hc.habInformer.GetStore().GetByKey(key)
```
LiliC: Should we not handle conforming the `PersistentVolumeClaim` as well? Or is there nothing that needs to be done, when for example Habitat resource gets deleted or updated?

pkg/apis/habitat/v1beta1/types.go
asymmetric: I gave it the same name it has in the `PersistentVolume` resource, because it [maps 1:1 with that field](
https://github.com/kinvolk/habitat-operator/tree/efe0e8fe74f3a93f550e606b9abad2bf81cd31b6/pkg/controller/stateful_sets.go#L206). My hope is that it will make it easier for people to understand how things tie together.

pkg/apis/habitat/v1beta1/types.go
asymmetric: What do you mean? Right now we only explicitly document that a field is optional, otherwise it's mandatory.

pkg/controller/controller.go
asymmetric: In theory so that the CM is deleted with the Habitat, but apparently that's [not working](#213)?

examples/persisted/README.md
asymmetric: I explicitly avoided to do that because it's a big topic and I would only make a poorer job at explaining it than the official docs do, which I linked. I anyway added some commits after your comment which should hopefully make things clearer.

BTW, it's not necessary that the PV has been created beforehand, if the cluster supports dynamic provisioning.

TL;DR: IMHO explaining how persistence works in Kubernetes is beside the scope of this README.

examples/persisted/README.md
asymmetric: OK, I'll add a link to [this](https://kubernetes.io/docs/concepts/storage/persistent-volumes/)

pkg/controller/controller.go
LiliC: That doesn't link to anything for me? 🤔

pkg/controller/controller.go
asymmetric: As mentioned in the [example README](https://github.com/habitat-sh/habitat-operator/blob/64e7f10e39792073294f891d4a70ebdaf3a1a51b/examples/persisted/README.md), we do nothing when they get deleted, because we want to prioritize data safety (which is also the default behavior with STS).

Do you think we should do something?

pkg/apis/habitat/v1beta1/types.go
LiliC: Cool, that makes sense!

pkg/apis/habitat/v1beta1/types.go
LiliC: Ok, just before in the docs.md file it was more clear at least to me, what is required. But if it makes sense to you 👍

examples/rbac/minikube.yml
```diff
@@ -0,0 +1,62 @@
...
+  - kind: ServiceAccount
+    name: default
+    namespace: kube-system
+  - kind: User
+    name: kube-proxy
+roleRef:
+  kind: ClusterRole
+  name: cluster-writer
+  apiGroup: rbac.authorization.k8s.io
+
```
LiliC: Line too much?

examples/rbac/minikube.yml
```diff
@@ -0,0 +1,62 @@
...
+apiVersion: rbac.authorization.k8s.io/v1beta1
+metadata:
+  name: cluster-writer
+rules:
+  - apiGroups: ["*"]
+    resources: ["*"]
+    verbs: ["*"]
+  - nonResourceURLs: ["*"]
+    verbs: ["*"]
+
```
LiliC: One line too much. :)

examples/rbac/minikube.yml
```diff
@@ -0,0 +1,62 @@
...
+    namespace: kube-system
+  - kind: User
+    name: kube-proxy
+roleRef:
+  kind: ClusterRole
+  name: cluster-writer
+  apiGroup: rbac.authorization.k8s.io
+
+---
+
```
LiliC: Remove line? :)

examples/persisted/README.md
```diff
@@ -0,0 +1,39 @@
...
+* familiarize yourself with how persistent storage [works in
+Kubernetes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/)
+* create a
+[`StorageClass`](https://kubernetes.io/docs/concepts/storage/storage-classes/)
+object with name `example-sc`
+* add the `spec.persistentStorage` key to the Habitat object's manifest
+* specify the `name` of the aforementioned `StorageClass` object under
+`spec.persistence.storageClassName` in the Habitat object's manifest
+
+An example `StorageClass` for clusters running on minikube is provided in
```
LiliC: Maybe note it doesn't work if it's vm-driver=none?

pkg/controller/controller.go
asymmetric: Fixed.

examples/rbac/minikube.yml
asymmetric: It's required/recommended to have a newline at the end of a file. My editor does it automatically.

Git complains if you don't have it.

examples/rbac/minikube.yml
asymmetric: I actually think that's not even needed, not sure why it's there in the first place. There's no `sd-build` `ServiceAccount` on my minikube cluster.

examples/persisted/README.md
asymmetric: I don't think anyone's gonna try it with that flag, considering it's not even documented ;)

examples/rbac/minikube.yml
asymmetric: Except that's not the end :rofl:

examples/persisted/README.md
LiliC: Think it is now. :D

examples/rbac/minikube.yml
asymmetric: Wrong position :joy: My comment referred to the bottom `ClusterRoleBinding`.

pkg/apis/habitat/v1beta1/types.go
asymmetric: AFAICT that's also how its' done in upstream.

PR #202, opened 2018-03-06
Reduce scope of RBAC authorization for E2E tests

asymmetric:
This PR:

* only gives namespaced permissions to the operator, since it doesn't need cluster-level ones
* simplifies the cleaning up after the test, because everything now resides under the `testing` namespaces


zeenix:
LGTM, except one nitpick on one of the commits: Don't see any reason to start a new paragraph for each sentence (but still better than the widespread habit of starting a new line for each sentence :-) ).

asymmetric:
I only break lines when I think it makes sense.

asymmetric:
As it is, this doesn't work, because the CRD needs to get created outside of the `testing` namespace, which the operator cannot do with the above rules.

Closing, as this is not particularly important.

PR #204, opened 2018-03-12
Improvements to .travis.yml

asymmetric:
* Stable go
* Extract k8s version to variable, to easily keep `kubectl` and minikube in sync

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ env:
 install: true
 
 go:
-  - 1.8.x
+  - stable
 
 git:
   depth: 1
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ sudo: required
 # This moves Kubernetes specific config files.
 env:
 - CHANGE_MINIKUBE_NONE_USER=true
-- K8S_VERSION=v1.8.0
+- K8S_VERSION=v1.9.0
 
 # Skip installation of dependencies since they are already in vendor.
 install: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,9 @@ sudo: required
 
 # This moves Kubernetes specific config files.
 env:
-- CHANGE_MINIKUBE_NONE_USER=true
-- K8S_VERSION=v1.9.0
+  global:
+    - CHANGE_MINIKUBE_NONE_USER=true
+    - K8S_VERSION=v1.9.0
 
 # Skip installation of dependencies since they are already in vendor.
 install: true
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -5,7 +5,7 @@ sudo: required
 # This moves Kubernetes specific config files.
 env:
 - CHANGE_MINIKUBE_NONE_USER=true
-- K8S_VERSION=v1.8.0
+- K8S_VERSION=v1.9.0
```
zeenix: This patch fixes a patch in the same PR so it should have been squashed into it before PR. It's no biggie, just for future reference. Could you squash it now?

.travis.yml
```diff
@@ -4,8 +4,9 @@ sudo: required
 
 # This moves Kubernetes specific config files.
 env:
-- CHANGE_MINIKUBE_NONE_USER=true
-- K8S_VERSION=v1.9.0
+  global:
+    - CHANGE_MINIKUBE_NONE_USER=true
+    - K8S_VERSION=v1.9.0
```
zeenix: Same comment here.

.travis.yml
asymmetric: I don't think so. In case someone decides that 2ce27aa63b9efc58969e090495f1f9a12734be8d is good while this is bad, they would only have to revert this patch.

.travis.yml
asymmetric: Same answer, this should be revertable independently of other changes.

zeenix: Maybe during the review but in the end they are one atomic change (at least the first two) so you should mention in the commit log that it's to be squashed before merge.

asymmetric: Well, even after merge, someone might disagree with the followup change and only want to revert that. Why shouldn't that be possible?

zeenix: They can edit 1-3 lines very easily. No need for all reverts to be a simple git command. :)

zeenix: but having a look again at the changes, i should have marked my comment as "big nitpick" :)

asymmetric: Yeah, I think it makes more sense to be consistent, regardless of how big the change would have to be.

.travis.yml
```diff
@@ -4,23 +4,25 @@ sudo: required
 
 # This moves Kubernetes specific config files.
 env:
-- CHANGE_MINIKUBE_NONE_USER=true
+  global:
+    - CHANGE_MINIKUBE_NONE_USER=true
+    - K8S_VERSION=v1.9.0
```
LiliC: hmmm... should we bump to 1.9.0 already, I remember there being some problems.

asymmetric: I think we can, at least on our tests. 1.10 is almost out, I see no reason not to.

PR #205, opened 2018-03-12
Code generator

asymmetric:
Adds a script to run the code generator under `hack/`.

The code generator itself must be installed separately, as explained in `update-codegen.sh`.


zeenix:
What kind of code this generates?

asymmetric:
See https://blog.openshift.com/kubernetes-deep-dive-code-generation-customresources/

It generates deepcopy functions for each of the types we define, since go has no generics.

asymmetric:
PTALA.

hack/update-codegen.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors.
```
LiliC: Not sure about this copyright? Did you mean Chef?

Makefile
```diff
@@ -40,4 +40,7 @@ update-version:
 	find examples helm -regex '.*\.ya?ml$$.bak' -type f \
 		-exec rm '{}' \;
 
+codegen:
+	CODEGEN_PKG=../../../k8s.io/code-generator hack/update-codegen.sh
```
LiliC: Can't you just do `CODEGEN_PKG=$GOPATH/src/k8s.io/code-generator`?

asymmetric: No, it has to be a relative path. :man_facepalming:

LiliC:
My only concern is the following:

I tried running this and it seems like I have an older version of the script as I didn't recently update the code-generator repo, so there was a difference when I ran it. I see two options:
1. Commit the script to the repo so we all run the same one.
2. Specify/remind in the README to update to latest version of the code-generator repo.

Up to you which one you choose. :)

asymmetric:
I would exclude 1. because the script gets updated somewhat regularly. I see 2 options:

* Run the `verify-codegen` script with by cloning the latest codegen in Travis, so that the build fails if versions differ
* Add the codegen as a git submodule to the repo (not sure I really want to deal with this one)

asymmetric:
OTOH, if there are no other problems, I think it would make sense to merge this now, so that I can use it in the other outstanding PRs. I would then add a separate ticket to make it more robust.

PR #207, opened 2018-03-13
Standardize optional fields

### LINKED ISSUE #186: Add +optional tag to optional fields
As recommended [here](https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#optional-vs-required).

asymmetric:
The `Group` field can now be set to the empty string `''`, which was previously not possible since it would default to the string `default`.

Additionally:

* Uses the `+optional` comment tag
* Makes all optional fields pointer types, unless their type has a zero-value of `nil`.
* Always uses `omitempty`.

Closes #186.


pkg/controller/controller.go
```diff
@@ -702,8 +702,8 @@ func (hc *HabitatController) newDeployment(h *habv1beta1.Habitat) (*appsv1beta1.
 	}
 
 	// Handle ring key, if one is specified.
-	if ringSecretName := h.Spec.Service.RingSecretName; ringSecretName != "" {
-		s, err := hc.config.KubernetesClientset.CoreV1().Secrets(apiv1.NamespaceDefault).Get(ringSecretName, metav1.GetOptions{})
+	if ringSecretName := h.Spec.Service.RingSecretName; ringSecretName != nil {
```
krnowak: You could probably have:

```
if ringSecretNamePtr := h.Spec.Service.RingSecretName; ringSecretNamePtr != nil {
    ringSecretName := *ringSecretNamePtr
```

That would limit the number of changed lines below. But I'm not insisting.

asymmetric: Yeah, I could... But if you're not insisting, I'll leave it like it is ;)

asymmetric:
Currently, passing `group: ''` or `group: ` both result in an invocation without the `--group` flag, so it seems unchanged.

asymmetric:
That's a good point. I think we should mention this in the CHANGELOG when this PR gets merged.

LiliC:
> That's a good point. I think we should mention this in the CHANGELOG when this PR gets merged.

Have you considered bumping up the `v1beta1` to `v1beta2` as well? Since it's sort of a breaking change?

asymmetric:
Sure, if that's the recommended approach. What does SemVer say about this?

pkg/apis/habitat/v1beta1/types.go
```diff
@@ -62,19 +62,20 @@ type HabitatState string
...
+	// +optional
+	Group *string `json:"group,omitempty"`
 	// Topology is the value of the --topology flag for the hab client.
 	Topology `json:"topology"`
 	// ConfigSecretName is the name of a Secret containing a Habitat service's config in TOML format.
 	// It will be mounted inside the pod as a file, and it will be used by Habitat to configure the service.
-	// Optional.
-	ConfigSecretName string `json:"configSecretName,omitempty"`
+	// +optional
+	ConfigSecretName *string `json:"configSecretName,omitempty"`
```
LiliC: 👍 :)

krnowak:
LFAD.

PR #208, opened 2018-03-13
Replace kinvolk with habitat

asymmetric:
* For the code, replace with "habitat-sh"
* For Docker images, with "habitat"


asymmetric:
I'll reopen against the fork.

PR #210, opened 2018-03-14
Run verify-codegen.sh on Travis

### LINKED ISSUE #206: Make code generation more reproducible
Since the [code generator](https://github.com/kubernetes/code-generator) we use in the `Makefile` is not part of this repo (and cannot be, since `dep` does not support non-Go dependencies), there's a chance that different versions of it produce different outputs.

In order to solve this, we could:

* Make Travis clone the codegen repo (maybe at a stable tag/commit) and run [verify-codegen.sh](https://github.com/kubernetes/code-generator/blob/master/hack/verify-codegen.sh), making Travis the "source of truth"
* Add the codegen repo as a git submodule, making this repo the "source of truth"

I prefer the first solution, as git submodules are kind of a UX nightmare for everyone involved.

asymmetric:
The `hack/verify-codegen.sh` script checks whether there are any pending codegen changes the committer has forgotten to commit.

Closes #206.


asymmetric:
Closing, it wants to merge on the wrong branch.

PR #211, opened 2018-03-14
Run verify-codegen.sh on CI

### LINKED ISSUE #206: Make code generation more reproducible

asymmetric:
The `hack/verify-codegen.sh` script checks whether there are any pending codegen changes the committer has forgotten to commit.

Closes #206.


hack/verify-codegen.sh
```diff
@@ -0,0 +1,40 @@
...
+"${SCRIPT_ROOT}/hack/update-codegen.sh"
+echo "checking against freshly generated codegen"
+
+ret=0
+
+git diff --quiet "*.deepcopy.go" || ret=$?
+
+if [[ $ret -eq 0 ]]
+then
+  echo "Up to date."
```
LiliC: Maybe something like this for more clarity:
> Deepcopy generated files up to date.

hack/verify-codegen.sh
```diff
@@ -0,0 +1,40 @@
...
+else
+  echo "Out of date. Please run hack/update-codegen.sh"
```
LiliC: And this something like:

> Deepcopy generated files out of date. Please run hack/update-codegen.sh

.travis.yml
```diff
@@ -25,13 +25,15 @@ before_script:
 - sudo minikube start --vm-driver=none --kubernetes-version=${K8S_VERSION} --extra-config=apiserver.Authorization.Mode=RBAC
 # Fix the kubectl context, as it's often stale.
 - minikube update-context
+# Clone the code generator repo
+- mkdir -p $GOPATH/src/k8s.io && git clone https://github.com/kubernetes/code-generator.git $GOPATH/src/k8s.io/code-generator
```
LiliC: Think this would be better to be part of the verify script, so if it that needs to get removed you don't need to chase down the dependencies for it.

.travis.yml
```diff
@@ -25,13 +25,15 @@ before_script:
...
 # Wait for Kubernetes to be up and ready.
 - JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}'; until kubectl get nodes -o jsonpath="$JSONPATH" 2>&1 | grep -q "Ready=True"; do sleep 1; done
-- make TAG=testing image
 
 script:
+- CODEGEN_PKG=../../../../src/k8s.io/code-generator hack/verify-codegen.sh
 - make test
-- make TESTIMAGE=habitat/habitat-operator:testing e2e
+- make TAG=testing image && make TESTIMAGE=habitat/habitat-operator:testing e2e
```
LiliC: Maybe just add `image` target to `e2e` instead, so it's run before, as it's needed always anyways.

hack/verify-codegen.sh
asymmetric: Just above it says "checking against freshly generated codegen", maybe that's enough?

.travis.yml
asymmetric: Interesting point.. I'll give it a try.

.travis.yml
asymmetric: I thought about this, but I think it's not a good idea: the `image` target is not built in a way that it knows when it doesn't have to perform any work. Which means that running `e2e` twice would build the Docker image twice.

Maybe we can refactor the `image` target to be idempotent, but I didn't think it was in scope for this PR.

asymmetric: I take that back, Docker actually uses the cache, so we can do what you said.

hack/verify-codegen.sh
LiliC: Nitpick alert! :)

Only reason I said was because currently shows up like this in the Travis logs:
> checking against freshly generated codegen
> Up to date.

And it doesn't seem connected, maybe just add a `:` to understand that those two lines of logs belong together.

https://travis-ci.org/habitat-sh/habitat-operator/builds/353421977#L523

.travis.yml
asymmetric: OK, here's the problem: `go build` is run every time `make e2e` is run, which doesn't seem right.

LiliC: Can you explain why doesn't it seem right?

asymmetric: Sorry, wasn't very clear: because it builds even when nothing has changed, so I end up waiting longer for no reason.

LiliC: Well it's mainly aimed to run on travis and since we don't do cache there should be fine...

asymmetric: I run it locally :)

LiliC: Why? :D

asymmetric: Because we have different setups on Travis and locally (running minikube with/without a VM), so some things behave differently.

LiliC: Fair enough! We could also have one target that is `make travis` as well, if I remember what this discussion was about anyways. :D

LiliC:
Is this ready for a re-review?

asymmetric:
Updated to:

* Run on CircleCI
* Pin the correct branch for the version of Kubernetes we're targeting, rather than `master

asymmetric:
PTALA, test was flaky on CircleCI, hopefully #233 will fix this.

hack/verify-codegen.sh
```diff
@@ -0,0 +1,40 @@
...
+# limitations under the License.
+
+# Based on https://github.com/kubernetes/code-generator/blob/893b4433a4ba929dd0dacebf7c8956682d7a5d5f/hack/verify-codegen.sh
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE}")/..
+SCRIPT_BASE=${SCRIPT_ROOT}/../..
```
krnowak: This line looks unnecessary - `SCRIPT_BASE` is not used anywhere.

asymmetric:
Fixed the buld by pinning minikube. The latest version has [issues](https://github.com/kubernetes/minikube/issues/2704) running on both Travis and CircleCI :man_facepalming:

asymmetric:
Sure, feel free to open an issue. I didn't change it much compared to the original, for simplicity's sake.

krnowak:
Filed as #238.

PR #212, opened 2018-03-15
Add habitat plan

elliott-davis:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,4 @@
 .glide/
 
 /habitat-operator
+results/
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ HUB :=
 GITHUB_ORG := habitat-sh
 DOCKER_ORG := habitat
 IMAGE := $(if $(HUB),$(HUB)/)$(DOCKER_ORG)/habitat-operator
+BIN_PATH := habitat-operator
 TAG := $(shell git describe --tags --always)
 TESTIMAGE :=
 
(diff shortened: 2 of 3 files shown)


zeenix:
@elliott-davis cool stuff but I think you meant to write "Add habitat-operator plan" instead in the titles. :) My first thought was "inception".

LiliC:
Sorry, I am sort of missing the point of adding this, at first glance at least. :) Is it so that you can run the operator with habitat in your k8s cluster?

asymmetric:
@elliott-davis I'm also not 100% sure about this, but it might just be that I'm not understanding your use-case.

Could you explain a bit the rationale?

elliott-davis:
@lili exactly! It's so we can leverage the habitat exporters and the
habitat builder service to manage the habitat operator container. I'm happy
to add that to the commit message also.

LiliC:
@elliott-davis One thing I do not understand is this part:

> It's so we can leverage the habitat exporters

What do you mean by that, how would the workflow look like? Thanks!

elliott-davis:
hey @LiliC - So I was envisioning it working like so:
1. Connect the plan file to Builder
2. Enable the export to DockerHub exporter
3. On subsequent changes to the operator, the build service will build and export the new operator to DockerHub

I was hoping that this could replace the `make image` target and streamline the Docker export

asymmetric:
@elliott-davis thanks! I see how this could be helpful, but OTOH i still think that the `image` target has a place, e.g. for local development.

Anyway, this LGTM.

PR #214, opened 2018-03-20
Update helm charts as per the repository move

zeenix:
NOTE: This PR is for gh-pages branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,23 +4,25 @@ sudo: required
 
 # This moves Kubernetes specific config files.
 env:
-- CHANGE_MINIKUBE_NONE_USER=true
+  global:
+    - CHANGE_MINIKUBE_NONE_USER=true
+    - K8S_VERSION=v1.9.0
 
 # Skip installation of dependencies since they are already in vendor.
 install: true
 
 go:
-  - 1.8.x
+  - stable
 
 git:
   depth: 1
 
 before_script:
 # Download kubectl, which is a requirement for using minikube.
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -4,23 +4,25 @@ sudo: required
 
 # This moves Kubernetes specific config files.
 env:
-- CHANGE_MINIKUBE_NONE_USER=true
```
LiliC: Why do you need a CI for this helm branch anyways?

helm/habitat-operator/values.yaml
```diff
@@ -1,7 +1,7 @@
 ## Habitat-operator image
 ##
 image:
-  repository: docker.io/kinvolk/habitat-operator
```
LiliC: Was this updated?

.travis.yml
zeenix: This is just merging changes from master. But since you asked, yeah we could remove this file afterwards.

helm/habitat-operator/values.yaml
zeenix: I am not sure, this is also just a merge/cherrypick from master. I assume all the modified link in the master branch where tested.

zeenix:
> Can't you just rebase this branch on master?

We discussed a while back. We cant have the same content as master in this branch and hence a simple rebase/merge is not appropriate. Most updates are actually mostly automated through the chart-for-latest.sh script in this branch.

asymmetric:
I see. Well, then either in this PR or in a following one we should remove files that are not relevant.

.travis.yml
LiliC: Yes, please!

zeenix:
@LiliC i can really do that. The `gh-pages` is what is served and not `zeenix/gh-pages` so I can really test against helm.

PR #217, opened 2018-03-23
Fix config example

LiliC:
The service name of the redis Habitat config example was wrong, so when
the user.toml path was formed the user.toml was placed in the wrong pkg
directory.


LiliC:
I checked the other examples and they all seem to have the correct Service name.

PR #224, opened 2018-03-28
Use relative path for `go build`

asymmetric:
This allows us to build the project in forks, since we don't hardcode
an absolute import path.


krnowak:
Travis failure seemed to be unrelated, restarted.

krnowak:
@asymmetric: Probably LFAD. Is this to have travis running builds on kinvolk repo working correctly?

asymmetric:
@krnowak yep.

PR #225, opened 2018-03-28
Run tests on CircleCI

asymmetric:
This PR enables the running of tests on CircleCI.

The main advantage is that, in case of a failing build, we can SSH into the machine.

I think we should run both CircleCI and Travis, evaluate the respective benefits, and then converge on one solution.

**NOTE**: ~I tried using the `CHANGE_MINIKUBE_NONE_USER=true` env variable, but it somehow got ignored. Therefore I resorted to manual `chown`/`chgrp`.~

**NOTE**: If https://github.com/habitat-sh/habitat-operator/pull/211 is merged, this needs to be updated to also run the `update-codegen.sh` script.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,39 @@
+version: 2
+jobs:
+  build:
+    machine: true
+    working_directory: ~/.go_workspace/src/github.com/habitat-sh/habitat-operator
+    steps:
+      - checkout
+      - run:
+          name: setup
+          environment:
+            K8S_VERSION: v1.9.0
+          command: |
+            curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/${K8S_VERSION}/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - checkout
+      - run:
+          name: setup
+          environment:
+            K8S_VERSION: v1.9.0
+          command: |
+            curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/${K8S_VERSION}/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+            curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
+            sudo minikube start --vm-driver=none --kubernetes-version=${K8S_VERSION} --extra-config=apiserver.Authorization.Mode=RBAC
```
krnowak: Use `sudo -E` for the `CHANGE_MINIKUBE_NONE_USER` env var to be forwarded to minikube's environment.

https://github.com/kubernetes/minikube/issues/1883#issuecomment-324689491

asymmetric:
PTALA.

PR #229, opened 2018-03-29
Make it possible to use sudo when calling docker

krnowak:
Docker on Fedora requires root to talk to it, so `make image` always
fails there currently. But let's not just hardcode `sudo` to fix it,
because other distros may be more convenient here by not requiring
root when talking to docker, so sudo here is pointless. With this
change, we can do `make SUDO=sudo image` on Fedora to make it work.


asymmetric:
Would `sudo make image` work?

krnowak:
It kinda would, but it is not what I want.

First, it would need to be `sudo -E make linux` otherwise `GOPATH` env var is not forwarded, so building the static binary will fail. Second, it produces a binary file owned by root in the source directory, which I don't want to have.

PR #239, opened 2018-04-11
Handle multiple CRD versions

### LINKED ISSUE #215: Generate client with code-generator
The [code generator](https://github.com/kubernetes/code-generator) can generate a client for us.

This will allow us to:

* Conform more easily to the expected structure of other controllers/operator
* Support multiple versions without having to manually rewrite/modify the existing client
* Write less code!

### LINKED ISSUE #216: Add habv1beta2 version
We decided to support the existing `Deployment`-based behavior in `habv1beta`, and to move the `StatefulSet`-based logic to `habv1beta2`.

This allows us to gracefully deprecate the old behavior without disrupting existing clusters.

Additionally, we decided to keep supporting `habv1beta1` for 2 releases,

See #215.

asymmetric:
This PR does several things.

## CustomVersion

There is currently no way to run multiple versions of a CRD. The `CustomVersion` field works around this by "overloading" the Habitat type.

All controllers receive all `Habitat` objects, but then decide which ones to skip based on the value (and presence) of the `CustomVersion` field.

## Code generator

This PR also reorganizes the `pkg` tree to match the expectations of the code-generator. The code-generator is used to generate client, listers and watchers.

In the future, we will be able to maintain multiple versions more easily thanks to this structure.

The use of the code-generator also implied changes in the controller, because the structure of the client, listers and watchers has changed.

## Backwards compatibility

The new field allows us to maintain backwards-compatibility with clusters that are already running `Deployment`-backed `Habitat` objects.

Closes #215.
Closes #216.


asymmetric:
Closing, wrong remote.

PR #244, opened 2018-04-17
vendor: update version of k8s.io repos to satisfy the dep gods

iaguis:
In the habitat-service-broker we want to vendor v0.1.13 of the
service-catalog which specifies kubernetes-1.9.1 as a constraint. This
conflicts with the habitat-operator.

There's no code change in vendored files.


asymmetric:
Why not the latest minor, i.e. `kubernetes-1.9.6`?

iaguis:
> EDIT: Get it, because you want to match the version in the SB. Wonder if a higher version would work, though.

I pushed a commit with `kubernetes-1.9.6` and @indradhanush tested it and it didn't work. I'll go back to the previous version.

asymmetric:
Can we close this?

iaguis:
Let's close this since it's not needed for now.

PR #249, opened 2018-04-23
helm release fixes

krnowak:
Fix the RBAC rules. Update the helm package maintainers list.

### DIFF
--- a/helm/habitat-operator/Chart.yaml
+++ b/helm/habitat-operator/Chart.yaml
@@ -2,8 +2,12 @@ apiVersion: v1
 description: A Kubernetes controller that manages Habitat services in a Kubernetes cluster.
 engine: gotpl
 maintainers:
-  - name: Zeeshan Ali
-    email: zeeshan@kinvolk.io
+  - name: Krzesimir Nowak
+    email: krzesimir@kinvolk.io
+  - name: Iago López Galeiras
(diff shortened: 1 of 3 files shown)


helm/habitat-operator/Chart.yaml
```diff
@@ -2,8 +2,12 @@ apiVersion: v1
 description: A Kubernetes controller that manages Habitat services in a Kubernetes cluster.
 engine: gotpl
 maintainers:
-  - name: Zeeshan Ali
-    email: zeeshan@kinvolk.io
+  - name: Krzesimir Nowak
+    email: krzesimir@kinvolk.io
+  - name: Iago López Galeiras
+    email: iago@okinvolk.io
```
asymmetric: typo

krnowak: Ah, so that's where this `o` went. ;)

asymmetric:
One more case in which the manifests under `examples/` and the ones under `test/e2e/` are out of sync :(

I really think we should find a better way here.

helm/habitat-operator/Chart.yaml
krnowak: Or should that be `iago@ocisly.io`?

krnowak:
Updated.

PR #257, opened 2018-04-25
CHANGELOG: Deprecate customVersion = v1beta1

asymmetric:
We follow the release schedule of Kubernetes, according to which
deprecated versions are dropped after 3 releases.

`v1beta2` was [introduced](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.8.md#workloads-api-appsv1beta2) in 1.8, thereby deprecating `v1beta1`, which
should therefore be dropped by 1.11.

 `customVersion = v1beta1` maps to a controller that uses `apps/v1beta1`,
 which won't be updated, and that must therefore be droppped.


CHANGELOG.md
```diff
@@ -17,6 +17,13 @@
 
 Please compare the `v1beta1` and `v1beta2` manifests of the standalone example in `examples/v1beta1/habitat.yml` and `examples/standalone/habitat.yml`, respectively, to compare the immediate differences between them. Please refer to `examples/persisted/habitat.yml` for an example of the persistence functionality.
 
+### Deprecations
+
+* `Habitat` Manifests that do not specify a `customVersion`, or that specify a
+`customVersion = v1beta1` are deprecated, and support for them will be removed
+when Kubernetes 1.11 is released. Please upgrade your manifests by setting
```
krnowak: I'm not sure I like the last sentence. Setting `customVersion = v1beta2` is not enough. You need to use `specv1beta2` field too. Maybe just drop this sentence or say something like "Please upgrade your manifests to the latest `customVersion`"?

asymmetric:
I think it's good to mark it as deprecated now, even if we should've done it before.

> Maybe we should always mark previous custom versions as automatically deprecated

How would you do that? Were you thinking of adding that to the README?

krnowak:
> How would you do that? Were you thinking of adding that to the README?

Yep. Just a kind of a statement or a policy.

asymmetric:
Added it to the design doc.

krnowak:
LFAD. Could you add the deprecation notice here too? https://github.com/habitat-sh/habitat-operator/releases/tag/v0.6.0

PR #258, opened 2018-04-25
Kubernetes 1.10

### LINKED ISSUE #171: Check what needs to be upgraded for client-go v6
There are some [changes](http://blog.kubernetes.io/2018/01/introducing-client-go-version-6.html) to be made. Off the top of my head:

* `v1` instead of `v1beta1` for Deployments
* getting rid of [newListWatchFromClientWithLabels](
https://github.com/kinvolk/habitat-operator/tree/3fc2a17f09fbaa9088efc07fd1914f00909458cb/pkg/controller/utils.go#L66) and use `NewFilteredSharedInformerFactory`.

### LINKED ISSUE #219: Upgrade to 1.10

asymmetric:
See #219 for things this changes.

Closes #219.
Closes #171.


.circleci/config.yml
```diff
@@ -8,19 +8,20 @@ jobs:
...
             sudo minikube config set WantReportErrorPrompt false
-            sudo -E minikube start --vm-driver=none --kubernetes-version=${K8S_VERSION} --extra-config=apiserver.Authorization.Mode=RBAC
+            # TODO: remove the --bootstrapper flag once this issue is solved: https://github.com/kubernetes/minikube/issues/2704
+            sudo -E minikube start --vm-driver=none --bootstrapper=localkube --kubernetes-version=${K8S_VERSION} --extra-config=apiserver.Authorization.Mode=RBAC
       - run:
           name: code-gen script
           environment:
             CODEGEN_PKG: ../../../../src/k8s.io/code-generator
-            CODEGEN_VERSION: release-1.9
+            CODEGEN_VERSION: kubernetes-1.10.0
```
krnowak: Shouldn't it be `release-1.10`?

Whatever outcome is (either a tag or a branch), we should keep docs in the update codegen script in sync.

.circleci/config.yml
```diff
@@ -46,7 +47,9 @@ jobs:
           name: print habitat object logs
           command: kubectl logs -lhabitat-operator=true --tail=100
           when: on_fail
-      - run:
-          name: print minikube logs
-          command: minikube logs
-          when: on_fail
+      # NOTE: this relies on journalctl, which is not present on the version of
```
krnowak: So how did it work before? Old minikube/k8s didn't require journalctl?

.circleci/config.yml
```diff
@@ -46,7 +47,9 @@ jobs:
...
+      # Ubuntu currenlty run by CircleCI
```
krnowak: currently

.circleci/config.yml
```diff
@@ -46,7 +47,9 @@ jobs:
...
+      # - run:
+          # - runame: print minikube logs
```
krnowak: runame rucommand and ruwhen?

.travis.yml
```diff
@@ -8,9 +8,9 @@ go_import_path: github.com/habitat-sh/habitat-operator
 env:
   global:
     - CHANGE_MINIKUBE_NONE_USER=true
-    - K8S_VERSION=v1.9.0
-    - MINIKUBE_VERSION=v0.25.2
-    - CODEGEN_VERSION=release-1.9
+    - K8S_VERSION=v1.10.0
+    - MINIKUBE_VERSION=v0.26.1
+    - CODEGEN_VERSION=kubernetes-1.10.0
```
krnowak: `release-1.10`?

pkg/controller/v1beta1/controller.go
```diff
@@ -167,11 +167,11 @@ func (hc *HabitatController) cacheHabitats() {
 }
 
 func (hc *HabitatController) cacheDeployments() {
-	source := newListWatchFromClientWithLabels(
+	source := cache.NewFilteredListWatchFromClient(
```
krnowak: Should we touch this code at all? `v1beta1` custom version is frozen.

pkg/controller/v1beta2/controller.go
```diff
@@ -548,11 +548,11 @@ func (hc *HabitatController) conform(key string) error {
 	}
 
 	// Create StatefulSet, if it doesn't already exist.
-	if _, err := hc.config.KubernetesClientset.AppsV1beta1().StatefulSets(h.Namespace).Create(sts); err != nil {
+	if _, err := hc.config.KubernetesClientset.AppsV1beta2().StatefulSets(h.Namespace).Create(sts); err != nil {
```
krnowak: This probably should be a part of v1beta3 custom version, no? But maybe releasing v1beta3 will not be required, as supposedly you should be able to create a v1beta1 statefulset and handle it as v1beta2 statefulset. Maybe same goes for caching and so on. Could you check it?

pkg/controller/v1beta1/controller.go
asymmetric: That's true, OTOH it should be a drop-in replacement. Not sure, should I remove?

.travis.yml
asymmetric: Same as above

pkg/controller/v1beta2/controller.go
asymmetric: So I've:

* Started a 1.10 cluster
* Run the operator (out of cluster) on master (meaning, sts v1beta1 controller)
* Created the `standalone` example
* Killed the operator
* Checked out this branch
* Started the operator
* Edited the Habitat manifest

The updates were picked up by the operator.

So the `v1beta1` resource is delivered to the `v1beta2` watcher. So we don't need to upgrade to `v1beta3` for this change.

.circleci/config.yml
krnowak: These ones: https://github.com/habitat-sh/habitat-operator/blob/3adc004dd5371dd6ae516ae494784f3a7a7f34ff/hack/update-codegen.sh#L20-L28

pkg/controller/v1beta1/controller.go
krnowak: Yeah, I'd just leave it alone.

asymmetric:
PTALA.

asymmetric:
Good catch! `git rebase --signoff master` did the trick.

PR #259, opened 2018-04-25
add channel to supervisor

elliott-davis:
[Habitat updater](https://github.com/habitat-sh/habitat-updater) is a service that runs inside of a k8's cluster to watch for changes in pods running habitat services and reconcile any updates from builder. It does this by querying the k8's pods api for anything with a `habitat=true` label. Then it queries the supervisors running in those pods for what services they are running. Once it has a list of services, it asks builder for the most recent version of those packages in the stable channel. This is undesirable if you want to run a version of your package from the dev/acceptance channel. 

This change will allow you to tell the Habitat supervisor what channel it should be pinned to with the `--channel` flag. The update service will have to be modified after this change to get the channel from the supervisor and query builder accordingly.

### DIFF
--- a/pkg/apis/habitat/v1beta1/types.go
+++ b/pkg/apis/habitat/v1beta1/types.go
@@ -140,6 +140,12 @@ type ServiceV1beta2 struct {
 	// Name is the name of the Habitat service that this Habitat object represents.
 	// This field is used to mount the user.toml file in the correct directory under /hab/user/ in the Pod.
 	Name string `json:"name"`
+	// Channel is the value of the --channel flag for the hab client.
+	// It can be used to track upstream packages in builder channels but will never be used directly by the supervisor.
+	// The should only be used in conjunction with the habitat updater https://github.com/habitat-sh/habitat-updater
+	// Defaults to `stable`.
+	// +optional
+	Channel *string `json:"channel,omitempty"`
 }
 
 type Bind struct {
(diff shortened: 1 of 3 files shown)


pkg/apis/habitat/v1beta1/types.go
```diff
@@ -140,6 +143,9 @@ type ServiceV1beta2 struct {
 	// Name is the name of the Habitat service that this Habitat object represents.
 	// This field is used to mount the user.toml file in the correct directory under /hab/user/ in the Pod.
 	Name string `json:"name"`
+	// Channel is the value of the --channel flag for the hab client.
```
krnowak: This documentation doesn't really tell much about the purpose of this field.

elliott-davis:
@krnowak yes we do have supervisor updates turned off so for the supervisor, the action is a no-op. However, the habitat-updater service can use the channel to ask builder if there are any new packages available, then tell kubernetes to orchestrate the upgrade.

krnowak:
LFAD overall I think, but I would like to hear a second opinion. @asymmetric: ^

We used to have a related issue and a PR for it - #126 and #128.

In the meantime, could you write a bit better commit message, please? :) Some explanation along the lines of habitat-updater being able to figure out which channel the service should follow on updates.

asymmetric:
> the habitat-updater service can use the channel to ask builder if there are any new packages available, then tell kubernetes to orchestrate the upgrade.

Could you explain how this works?

How I understand it:

* The habitat-updater runs outside the cluster
* It uses the value of the channel flag to know what channel each service is running on
* It asks the builder if there are updates for that channel
* If so, it does whatever it needs to do to get them deployed to the cluster

Is that correct? Something along these lines could be a useful description for the PR, cause we are not very familiar with these other pieces that are now building on top of the operator :)

elliott-davis:
@asymmetric updated!

Issue #185, opened 2018-02-07
Use Schema for validation instead of ad-hoc code


asymmetric:
See [this](https://kubernetes.io/docs/tasks/access-kubernetes-api/extend-api-custom-resource-definitions/#validation) and [an example](https://github.com/kubernetes/sample-controller/blob/master/artifacts/examples/crd.yaml).

LiliC:
This would be great to have! 👍

asymmetric:
[Requires](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.8.md#custom-resource-definitions-crds) >= 1.8.

asymmetric:
Actually, it was alpha on 1.8, so we cannot count on it being turned on by default until 1.9, meaning it requires `>= 1.9`.

clintoncwolfe:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #267, opened 2018-05-02
Broadcast events

### LINKED ISSUE #221: Broadcast events
Events are displayed when doing `kubectl describe`, and are a useful way of seeing what happened historically to the object and its sub-resources, without having to scan the operator logs.

They can also be displayed with `kubectl get events`.

Events we might emit:

* `SuccessfulCreate` for the `StatefulSet`, `ConfigMap`, `Secret`, `PV`, `PVC` (the STS controller only does it for `Pod`s though)
* Failure for any of the above
* Habitat synced
* Validation failures (this should be removed once #185 is implemented)

asymmetric:
Emits an event when:

* A `Habitat` fails to validate (will have to be removed when #185 is implemented, because an invalid `Habitat` shouldn't even be creatable in the first place)
* One of the sub-resources fails to be created
* A `Habitat` is synced successfully (not sure we **really** need this)

To test: create a `Habitat` and type `kubectl describe hab`

Closes #221.

### DIFF
--- a/pkg/controller/v1beta2/controller.go
+++ b/pkg/controller/v1beta2/controller.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	habv1beta1 "github.com/habitat-sh/habitat-operator/pkg/apis/habitat/v1beta1"
+	habscheme "github.com/habitat-sh/habitat-operator/pkg/client/clientset/versioned/scheme"
 	habinformers "github.com/habitat-sh/habitat-operator/pkg/client/informers/externalversions"
 
 	"github.com/go-kit/kit/log"
@@ -35,8 +36,11 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	kubeinformers "k8s.io/client-go/informers"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	typedcorev1 "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 )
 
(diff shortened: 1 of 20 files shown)


asymmetric:
OK, removed the sync event, PTALA.

pkg/controller/v1beta2/controller.go
```diff
@@ -556,16 +594,19 @@ func (hc *HabitatController) conform(key string) error {
 				return err
 			}
 		} else {
+			hc.recorder.Event(h, apiv1.EventTypeWarning, stsFailed, messageStsCreated)
```
iaguis: Shouldn't this be `messageStsFailed`?

pkg/controller/v1beta2/controller.go
```diff
@@ -450,9 +484,11 @@ func (hc *HabitatController) handleConfigMap(h *habv1beta1.Habitat) error {
 			return err
 		}
 
-		level.Info(hc.logger).Log("msg", "updated peer IP in ConfigMap", "name", cm.Name, "ip", leaderIP)
+		level.Info(hc.logger).Log("msg", messagePeerIPUpdated, "name", cm.Name, "ip", leaderIP)
+		hc.recorder.Event(h, apiv1.EventTypeNormal, cmUpdated, messagePeerIPUpdated)
 	} else {
-		level.Info(hc.logger).Log("msg", "created peer IP ConfigMap", "name", cm.Name, "ip", leaderIP)
+		level.Info(hc.logger).Log("msg", messageCMCreated, "name", cm.Name, "ip", leaderIP)
+		hc.recorder.Event(h, apiv1.EventTypeNormal, cmUpdated, messagePeerIPAdded)
```
iaguis: Shouldn't this be a cm created event?

pkg/controller/v1beta2/controller.go
```diff
@@ -405,12 +437,14 @@ func (hc *HabitatController) handleConfigMap(h *habv1beta1.Habitat) error {
...
-			level.Debug(hc.logger).Log("msg", "removed peer IP from ConfigMap", "name", newCM.Name)
+			level.Debug(hc.logger).Log("msg", messagePeerIPRemoved, "name", newCM.Name)
+			hc.recorder.Event(h, apiv1.EventTypeNormal, cmUpdated, messagePeerIPRemoved)
 
 			return nil
 		}
 
-		level.Info(hc.logger).Log("msg", "created peer IP ConfigMap", "name", cm.Name)
+		level.Info(hc.logger).Log("msg", messageCMCreated, "name", cm.Name)
+		hc.recorder.Event(h, apiv1.EventTypeNormal, cmUpdated, messagePeerIPAdded)
```
iaguis: Shouldn't this be a cm created event?

iaguis:
I guess you need something like

```
...
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - get
  - update
  - patch
  - create
```

In the operator RBAC rules

PR #271, opened 2018-05-09
Remove dependency on pflag library

### LINKED ISSUE #78: Switch to pflag in test

asymmetric:
The only advantage was that it allowed us to define a shorthand (`-v`)
for the `--verbose` option, but we barely used that flag anyway, so the
additional overhead wasn't worth it.

Closes #78.
Supersedes and closes #170.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -428,6 +428,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "99212a1cfe34e275e246d205a124895d753180eef17ca71592ba2a63348548f4"
+  inputs-digest = "4756026b77d2dbae76c2ebaeb6c37f9c142cfc8c95d52a484b1325fce948f9cc"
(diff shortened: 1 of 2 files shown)


iaguis:
Should this close #170 too?

PR #274, opened 2018-05-14
Mark item done after validity check

asymmetric:
This PR improves two things about the way in which we handle the queue:

* Only mark items as done after we've checked them for validity (i.e. checked that they are strings)
* Call `Forget` for invalid items, so that we don't get them in the queue again.


iaguis:
#122 is already closed?

PR #275, opened 2018-05-14
controller: print custom version string instead of pointer

iaguis:
Otherwise the error message will include a memory address which is not
very useful.


asymmetric:
You're supposed to go check the memory address yourself, DUH!

LGTM.

iaguis:
> You're supposed to go check the memory address yourself, DUH!

We can implement a `--hard-mode` flag to go back to the previous behavior :D

PR #277, opened 2018-05-14
Dockerfile: start from scratch

### LINKED ISSUE #276: Use multi-stage Docker images
To reduce image size.

See [this](https://docs.docker.com/develop/develop-images/multistage-build/).

iaguis:
So we only have the static binary in the image and avoid including
alpine data.

This reduces the image size from 25.7MB to 21.7MB.

Closes #276

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,4 +2,4 @@ FROM scratch
 
 COPY habitat-operator /habitat-operator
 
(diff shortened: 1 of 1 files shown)


asymmetric:
See [this](https://github.com/golang/glog/blob/23def4e6c14b4da8ac2ed8007337bc5eb5007998/glog.go#L38).

PR #283, opened 2018-05-17
Add ServiceAccountName support to operator

elliott-davis:
When running services under the operator that need a service
account it is impossible to specify one given the current service
definition. This is useful for services running in pods that need to communicate with the kubernetes API with RBAC mode enabled


iaguis:
Sorry for the late reply.

It seems your code generator is old and it regenerated some files, that's why the tests fail.

I think @asymmetric and @krnowak had some ongoing discussions on where in the struct to put new fields. They're off until Monday, maybe we can sync about this then.

asymmetric:
I think you ran the code-generator at a version that's either too new or too old. Could you check out the `kubernetes-1.10.0` tag on the code-genearator and run it again?

Also, could you explain a usecase for this? AFAIU, you don't need to care about `ServiceAccount`s for your `Pod`s, unless the stuff running in them needs to talk to the APIServer, as they would default to the `default` SA.

Is this related to the work on the builder?

elliott-davis:
Ah I think I have an old version of the code generator.

I was trying to run the habitat-updater service inside GKE last week and
realized I couldn't because of rbac. After fiddling with it for a bit I saw
that is getting the default service account but it right have the right
priceless and I didn't want to grant it extra permissions.

asymmetric:
I guess the problem with running the operator on GKE with RBAC enabled is that we cannot create `ClusterRole`s. This should be fixed IMO by the person in charge of user authorizations.

I'm not sure we need support for custom `ServiceAccount`s once that's taken care of. WDYT?

elliott-davis:
I think both are true. An admin will need to create the role but we should
be able to tell a service what that role is, which is the goal of this pr.
Even the operator needs to know what it's service role is in its deployment
yaml

asymmetric:
I think two things are being conflated here. The operator needs its own SA (because it needs to do special tihngs with the API), and that's taken care of by `examples/rbac/rbac.yml`.

`Pod`s on the other hand most likely won't need to talk to the API, so they can just keep on using the `default` SA.

> we should be able to tell a service what that role is

Do we though? Are there usecases where a service running in a `Pod` has needed to talk to the API server?

> Even the operator needs to know what it's service role is in its deployment yaml

Sure, but that doesn't require changes to the `Habitat` type.

elliott-davis:
I this particular case, the service running inside the pod needs to talk to
the kubernetes API. See https://github.com/habitat-sh/habitat-updater

The updater is modifying the Habitat resource to update the image. When
this is run in an rbac env it barfs because that pod doesn't have
permission to patch resources, list other pods, etc

asymmetric:
@elliott-davis could you re-run the code-gen with the right tag?

The other missing thing would be an example using this key.

After that, I think we should be good to go :)

asymmetric:
@iaguis What was the usecase you mentioned where it would make sense to add the SA field?

iaguis:
If you have some application running in your cluster that's protected by RBAC and you want to access it from a service deployed with the Habitat Operator, you will need a Service Account that gives access to the RBAC-protected application.

For example, if you want a service that deploys Helm Charts, you probably need to pass the `tiller` Service Account to the service.

iaguis:
> RBAC-protected application

Although this means the application needs to be accessible through some kind of k8s API so maybe it's not the use case you're looking for...

examples/habitat-updater/habitat.yml
```diff
@@ -0,0 +1,12 @@
+apiVersion: habitat.sh/v1beta1
```
krnowak: Thanks for the examples! But I would like to have a readme for this example, where it says something about the service account field, and what is the use case. The yaml files are an example of the use case.

Just see the docs readmes in other examples.

Also, rename the example directory from `habitat-updater` to `service-account` or something.

examples/habitat-updater/habitat.yml
```diff
@@ -0,0 +1,12 @@
...
+kind: Habitat
+metadata:
+  name: habitat-updater
```
krnowak: Please add the following under `metadata`:

```
labels:
  source: operator-example
  app: service-account-habitat
```

krnowak (dismissed):
Looks fine, but I'd like to have some changes in the example.

clintoncwolfe:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #284, opened 2018-05-21
Use labels on examples

### LINKED ISSUE #154: Use labels on examples
I think we need to improve the naming of our example Habitats.

At the moment, we sometimes use `example-x-y-z`, which is fine for some cases (`example-standalone-habitat`), but then we also use `go` in other cases.

I tihnk from a discoverability/UX point of view, using labels would free us to then use the name that makes the most sense, but still letting the user clearly know that the Habitat is a) from the operator, b) an example.

kosyfrances:
Fixes #154 

We need to improve the naming of our example Habitats.
At the moment, we sometimes use example-x-y-z.
This is fine in some cases (example-standalone-habitat), but then we also use "go" in other cases.
From a discoverability/UX point of view, using labels would free us to use the name that makes the most sense,
but still letting the user clearly know that the Habitat is from the operator and is an example.

### DIFF
--- a/examples/bind-config/habitat.yml
+++ b/examples/bind-config/habitat.yml
@@ -13,6 +13,9 @@ apiVersion: habitat.sh/v1beta1
 kind: Habitat
 metadata:
   name: example-bind-configured-db
+  labels:
+    source: operator-example
+    app: bind-configured-db
 customVersion: v1beta2
 spec:
   v1beta2:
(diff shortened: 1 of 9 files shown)


krnowak:
Wouldn't [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) suit this purpose better?

kosyfrances:
@krnowak are you suggesting we use annotations instead. If we do not need the extra things labels have to offer, yes we can use annotations.

kosyfrances:
@asymmetric what say ye?

asymmetric:
It's conceivable that users could want to query this, so I think we should stick to labels.

PR #285, opened 2018-05-23
Support double signal handling

### LINKED ISSUE #220: Support double signal handling
See [this](https://github.com/kubernetes/sample-controller/blob/master/pkg/signals/signal.go).

This allows us to exit immediately if we receive two signals, in addition to exiting gracefully if we only receive one.

kosyfrances:
Fixes #220

### DIFF
--- a/cmd/habitat-operator/main.go
+++ b/cmd/habitat-operator/main.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"os/signal"
 	"runtime"
+	"sync"
 	"syscall"
 	"time"
 
@@ -90,7 +91,9 @@ func run() int {
 	}
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
-	defer cancelFunc()
+
+	var wg sync.WaitGroup
+	wg.Add(2)
 
 	cSets := Clientsets{
 		KubeClientset:          kubeClientset,
@@ -98,31 +101,40 @@ func run() int {
 		ApiextensionsClientset: apiextensionsClientset,
 	}
 
-	if err := v1beta1(ctx, cSets, logger); err != nil {
+	if err := v1beta1(ctx, &wg, cSets, logger); err != nil {
 		level.Error(logger).Log("msg", err)
 		return 1
 	}
 
-	if err := v1beta2(ctx, cSets, logger); err != nil {
+	if err := v1beta2(ctx, &wg, cSets, logger); err != nil {
 		level.Error(logger).Log("msg", err)
 		return 1
 	}
 
-	term := make(chan os.Signal)
+	term := make(chan os.Signal, 2)
 	// Relay these signals to the `term` channel.
 	signal.Notify(term, syscall.SIGINT, syscall.SIGTERM)
 
-	select {
-	case <-term:
-		level.Info(logger).Log("msg", "received SIGTERM, exiting gracefully...")
-	case <-ctx.Done():
-		level.Info(logger).Log("msg", "context channel closed, exiting")
-	}
+	go func() {
+		<-term
+		level.Info(logger).Log("msg", "received termination signal, exiting gracefully...")
+		cancelFunc()
+
+		<-term
+		os.Exit(1)
+	}()
+
+	<-ctx.Done()
+
+	// Block until the WaitGroup counter is zero
+	wg.Wait()
+
+	level.Info(logger).Log("msg", "controllers stopped, exiting")
 
 	return 0
 }
 
-func v1beta1(ctx context.Context, cSets Clientsets, logger log.Logger) error {
+func v1beta1(ctx context.Context, wg *sync.WaitGroup, cSets Clientsets, logger log.Logger) error {
 	// Create Habitat CRD.
 	_, err := habv1beta1controller.CreateCRD(cSets.ApiextensionsClientset)
 	if err != nil {
@@ -144,12 +156,15 @@ func v1beta1(ctx context.Context, cSets Clientsets, logger log.Logger) error {
 		return err
 	}
 
-	go controller.Run(runtime.NumCPU(), ctx)
+	go func() {
+		controller.Run(ctx, runtime.NumCPU())
+		wg.Done()
+	}()
 
 	return nil
 }
 
-func v1beta2(ctx context.Context, cSets Clientsets, logger log.Logger) error {
+func v1beta2(ctx context.Context, wg *sync.WaitGroup, cSets Clientsets, logger log.Logger) error {
 	// Create Habitat CRD.
 	_, err := habv1beta2controller.CreateCRD(cSets.ApiextensionsClientset)
 	if err != nil {
@@ -180,10 +195,24 @@ func v1beta2(ctx context.Context, cSets Clientsets, logger log.Logger) error {
 		return err
 	}
 
-	go kubeInformerFactory.Start(ctx.Done())
-	go habInformerFactory.Start(ctx.Done())
+	var factoriesWg sync.WaitGroup
+	factoriesWg.Add(2)
 
-	go controller.Run(runtime.NumCPU(), ctx)
+	go func() {
+		kubeInformerFactory.Start(ctx.Done())
+		factoriesWg.Done()
+	}()
+
+	go func() {
+		habInformerFactory.Start(ctx.Done())
+		factoriesWg.Done()
+	}()
+
+	go func() {
+		controller.Run(ctx, runtime.NumCPU())
+		factoriesWg.Wait()
+		wg.Done()
+	}()
 
 	return nil
 }
--- a/pkg/controller/v1beta1/controller.go
+++ b/pkg/controller/v1beta1/controller.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"reflect"
 	"regexp"
+	"sync"
 	"time"
 
 	"github.com/go-kit/kit/log"
@@ -106,20 +107,31 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
 }
 
 // Run starts a Habitat resource controller.
-func (hc *HabitatController) Run(workers int, ctx context.Context) error {
-	// Make sure the work queue is shutdown which will trigger workers to end.
-	defer hc.queue.ShutDown()
-
+func (hc *HabitatController) Run(ctx context.Context, workers int) error {
 	level.Info(hc.logger).Log("msg", "Watching Habitat objects")
+
+	var wg sync.WaitGroup
+	wg.Add(4 + workers)
 
 	hc.cacheHabitats()
 	hc.cacheDeployments()
 	hc.cacheConfigMaps()
-	hc.watchPods(ctx)
+	hc.watchPods(ctx, &wg)
 
-	go hc.habInformer.Run(ctx.Done())
-	go hc.deployInformer.Run(ctx.Done())
-	go hc.cmInformer.Run(ctx.Done())
+	go func() {
+		hc.habInformer.Run(ctx.Done())
+		wg.Done()
+	}()
+
+	go func() {
+		hc.deployInformer.Run(ctx.Done())
+		wg.Done()
+	}()
+
+	go func() {
+		hc.cmInformer.Run(ctx.Done())
+		wg.Done()
+	}()
 
 	// Wait for caches to be synced before starting workers.
 	if !cache.WaitForCacheSync(ctx.Done(), hc.habInformerSynced, hc.deployInformerSynced, hc.cmInformerSynced) {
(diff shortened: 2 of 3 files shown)


cmd/habitat-operator/main.go
```diff
@@ -112,12 +112,14 @@ func run() int {
...
 
-	select {
-	case <-term:
+	go func() {
+		<-term
 		level.Info(logger).Log("msg", "received SIGTERM, exiting gracefully...")
-	case <-ctx.Done():
-		level.Info(logger).Log("msg", "context channel closed, exiting")
-	}
+		cancelFunc()
```
krnowak: I think that after calling `cancelFunc`, we should do `<- term` again followed by `os.Exit(1)` or something. So the second ctrl-c will quit the operator immediately.

krnowak: Also, would be nice to figure out with calling `cancelFunc` twice won't give us problems. First call is after getting SIGTERM (or SIGINT) and the second call is deferred.

kosyfrances: After the first call, subsequent calls to a `cancelFunc` will do nothing.

asymmetric: True, see [docs](https://golang.org/pkg/context/#CancelFunc).

cmd/habitat-operator/main.go
```diff
@@ -112,12 +112,17 @@ func run() int {
 	// Relay these signals to the `term` channel.
 	signal.Notify(term, syscall.SIGINT, syscall.SIGTERM)
 
-	select {
-	case <-term:
+	go func() {
+		<-term
 		level.Info(logger).Log("msg", "received SIGTERM, exiting gracefully...")
```
asymmetric: Could also be `SIGINT`, so I'd remove mention of the specific signal and just say: "received termination signal, exiting gracefully..."

cmd/habitat-operator/main.go
```diff
@@ -112,12 +112,17 @@ func run() int {
...
-		level.Info(logger).Log("msg", "context channel closed, exiting")
-	}
+		cancelFunc()
+
+		<-term
+		os.Exit(1)
+	}()
+
+	<-ctx.Done()
+	level.Info(logger).Log("msg", "context channel closed, exiting")
```
asymmetric: This message is now redundant when exiting because of a signal. Are there cases when `ctx` gets closed and it's not because of signals?

I don't think so, so this message can probably go.

asymmetric:
Great writeup @krnowak!

kosyfrances:
@krnowak It's ready for another review

krnowak:
One thing I missed. The `defer hc.queue.Shutdown()` will likely need to stop being a `defer` and should be put between `<-ctx.Done()` and `wg.Wait()`. Otherwise it will probably deadlock…

krnowak:
Another thing - likely all the `wg.Done()` calls should be deferred, but `wg.Wait()` calls - not.

kosyfrances:
@krnowak Another one :)

cmd/habitat-operator/main.go
```diff
@@ -90,39 +91,50 @@ func run() int {
 	}
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
-	defer cancelFunc()
+
+	var wg sync.WaitGroup
+	wg.Add(4)
```
krnowak: I'd prefer this number to remain 2. This is because we pass it twice. I understand that you made it 4, because `v1beta2` function spawns not one goroutine, but three. Still, I would prefer to keep the number of spawned goroutines in the function as an implementation detail. That way it can be easier to reason about the code, like "we create a waitgroup with a counter set to 2 and pass it to two functions, so if I add a third function (like `v1beta3`), I need to bump this number to 3".

Right now the counter is 4, but we pass it to only two functions, so it may left the reader wondering why is that (do both functions decrement the counter twice?) and to be able to answer that, the reader needs to check both `v1beta1` and `v1beta2` functions to see what they do. Another problem - you need to remember to bump this number whenever you add another goroutine in either function, which is easy to forget, because the counter is not really in the vicinity of a body of either of the functions.

So, for the `v1beta2` function, I posted a possible solution at https://github.com/habitat-sh/habitat-operator/pull/285#pullrequestreview-125142243 (see the last snippet there, with the `factoriesWg` thing).

cmd/habitat-operator/main.go
```diff
@@ -144,12 +156,15 @@ func v1beta1(ctx context.Context, cSets Clientsets, logger log.Logger) error {
 		return err
 	}
 
-	go controller.Run(runtime.NumCPU(), ctx)
+	go func() {
+		go controller.Run(runtime.NumCPU(), ctx)
```
krnowak: While you are at it, could please swap the order of parameters here? `context.Context` should be the first parameter of a function. I suppose this is some kind of an expectation from static analysis tools or something.

Others could say that passing `ctx` as a first parameter is idiomatic.

cmd/habitat-operator/main.go
```diff
@@ -180,10 +195,20 @@ func v1beta2(ctx context.Context, cSets Clientsets, logger log.Logger) error {
...
+	}()
+
+	go func() {
+		go habInformerFactory.Start(ctx.Done())
+		wg.Done()
+	}()
 
-	go controller.Run(runtime.NumCPU(), ctx)
+	go func() {
+		go controller.Run(runtime.NumCPU(), ctx)
```
krnowak: Same here about swapping the order of parameters.

pkg/controller/v1beta1/controller.go
```diff
@@ -232,7 +253,16 @@ func (hc *HabitatController) watchPods(ctx context.Context) {
 		DeleteFunc: hc.handlePodDelete,
 	})
 
-	go c.Run(ctx.Done())
+	var wg sync.WaitGroup
```
krnowak: Hm, this looks wrong. It looks like an elaborate way of saying `c.Run(ctx.Done())`. ;)

What I think needs to be done there is that the `watchPods` function should receive waitgroup as a parameter, instead of creating it and it should not call `wg.Wait()` on it. Because right now, `watchPods` function will block, while it was meant to be a nonblocking function that spawns a goroutine.

pkg/controller/v1beta1/controller.go
```diff
@@ -107,19 +108,30 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
...
 	hc.cacheHabitats()
 	hc.cacheDeployments()
 	hc.cacheConfigMaps()
 	hc.watchPods(ctx)
 
-	go hc.habInformer.Run(ctx.Done())
-	go hc.deployInformer.Run(ctx.Done())
-	go hc.cmInformer.Run(ctx.Done())
+	var wg sync.WaitGroup
+	wg.Add(3 + workers)
```
krnowak: I think this will need to be bumped to 4, if we start passing it to `hc.watchPods` too.

pkg/controller/v1beta2/controller.go
```diff
@@ -152,19 +153,30 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
...
 	hc.cacheHabitats()
 	hc.cacheStatefulSets()
 	hc.cacheConfigMaps()
 	hc.watchPods(ctx)
 
-	go hc.habInformer.Run(ctx.Done())
-	go hc.stsInformer.Run(ctx.Done())
-	go hc.cmInformer.Run(ctx.Done())
+	var wg sync.WaitGroup
+	wg.Add(3 + workers)
```
krnowak: Same here about bumping to 4 and `hc.watchPods`.

pkg/controller/v1beta2/controller.go
```diff
@@ -230,7 +251,13 @@ func (hc *HabitatController) watchPods(ctx context.Context) {
 		DeleteFunc: hc.handlePodDelete,
 	})
 
-	go c.Run(ctx.Done())
+	var wg sync.WaitGroup
```
krnowak: Same here about receiving waitgroup as a parameter instead of creating a new one and waiting on it.

cmd/habitat-operator/main.go
krnowak: Oh, and drop the `go` here. We don't want to spawn a goroutine inside the goroutine.

cmd/habitat-operator/main.go
```diff
@@ -180,10 +195,20 @@ func v1beta2(ctx context.Context, cSets Clientsets, logger log.Logger) error {
 		return err
 	}
 
-	go kubeInformerFactory.Start(ctx.Done())
-	go habInformerFactory.Start(ctx.Done())
+	go func() {
+		go kubeInformerFactory.Start(ctx.Done())
```
krnowak: Drop `go`.

cmd/habitat-operator/main.go
```diff
@@ -180,10 +195,20 @@ func v1beta2(ctx context.Context, cSets Clientsets, logger log.Logger) error {
...
+		wg.Done()
+	}()
+
+	go func() {
+		go habInformerFactory.Start(ctx.Done())
```
krnowak: Drop `go`.

cmd/habitat-operator/main.go
krnowak: Drop `go` too.

pkg/controller/v1beta1/controller.go
```diff
@@ -107,19 +108,30 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
...
+
+	go func() {
+		go hc.habInformer.Run(ctx.Done())
```
krnowak: Drop `go`.

pkg/controller/v1beta1/controller.go
```diff
@@ -107,19 +108,30 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
...
+		wg.Done()
+	}()
+
+	go func() {
+		go hc.deployInformer.Run(ctx.Done())
```
krnowak: Drop `go`.

pkg/controller/v1beta1/controller.go
```diff
@@ -107,19 +108,30 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
...
+		wg.Done()
+	}()
+
+	go func() {
+		go hc.cmInformer.Run(ctx.Done())
```
krnowak: Drop `go`.

pkg/controller/v1beta1/controller.go
```diff
@@ -131,12 +143,21 @@ func (hc *HabitatController) Run(workers int, ctx context.Context) error {
 	// failed job, it will be restarted after a delay of 1 second.
 	for i := 0; i < workers; i++ {
 		level.Debug(hc.logger).Log("msg", "Starting worker", "id", i)
-		go wait.Until(hc.worker, time.Second, ctx.Done())
+		go func() {
+			go wait.Until(hc.worker, time.Second, ctx.Done())
```
krnowak: Drop `go`.

pkg/controller/v1beta1/controller.go
```diff
@@ -232,7 +253,16 @@ func (hc *HabitatController) watchPods(ctx context.Context) {
 		DeleteFunc: hc.handlePodDelete,
 	})
 
-	go c.Run(ctx.Done())
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		go c.Run(ctx.Done())
```
krnowak: Drop `go`.

pkg/controller/v1beta2/controller.go
```diff
@@ -152,19 +153,30 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
...
 	hc.watchPods(ctx)
 
-	go hc.habInformer.Run(ctx.Done())
-	go hc.stsInformer.Run(ctx.Done())
-	go hc.cmInformer.Run(ctx.Done())
+	var wg sync.WaitGroup
+	wg.Add(3 + workers)
+
+	go func() {
+		go hc.habInformer.Run(ctx.Done())
```
krnowak: Drop `go`.

pkg/controller/v1beta2/controller.go
```diff
@@ -152,19 +153,30 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
...
+		wg.Done()
+	}()
+
+	go func() {
+		go hc.stsInformer.Run(ctx.Done())
```
krnowak: Drop `go`.

pkg/controller/v1beta2/controller.go
```diff
@@ -152,19 +153,30 @@ func New(config Config, logger log.Logger) (*HabitatController, error) {
...
+		wg.Done()
+	}()
+
+	go func() {
+		go hc.cmInformer.Run(ctx.Done())
```
krnowak: Drop `go`.

pkg/controller/v1beta2/controller.go
```diff
@@ -176,12 +188,21 @@ func (hc *HabitatController) Run(workers int, ctx context.Context) error {
 	// failed job, it will be restarted after a delay of 1 second.
 	for i := 0; i < workers; i++ {
 		level.Debug(hc.logger).Log("msg", "Starting worker", "id", i)
-		go wait.Until(hc.worker, time.Second, ctx.Done())
+		go func() {
+			go wait.Until(hc.worker, time.Second, ctx.Done())
```
krnowak: Drop `go`.

pkg/controller/v1beta2/controller.go
```diff
@@ -230,7 +251,13 @@ func (hc *HabitatController) watchPods(ctx context.Context) {
 		DeleteFunc: hc.handlePodDelete,
 	})
 
-	go c.Run(ctx.Done())
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	go func() {
+		go c.Run(ctx.Done())
```
krnowak: Drop `go`.

kosyfrances:
@krnowak Another one 💯

krnowak:
LFAD.

PR #302, opened 2018-06-20
Run E2E tests on GCE

### LINKED ISSUE #282: Run E2E tests on GCE
See #266.

kosyfrances:
This PR does the following:

- Downloads gcloud
-  Authenticates using secrets
-  Creates cluster on GKE
- Pushes docker image to GCR
-  Runs e2e test on the GKE cluster, using the image from GCR

Fixes #282 

To be done: Document steps for users


.circleci/config.yml
```diff
@@ -3,6 +3,8 @@ jobs:
   build:
     machine: true
     working_directory: ~/.go_workspace/src/github.com/habitat-sh/habitat-operator
+    environment:
+      IMAGE_NAME: habitat/habitat-operator
```
krnowak: Would setting `IMAGE_NAME` to `eu.gcr.io/$GCLOUD_PROJECT_ID/$IMAGE_NAME-$CIRCLE_BUILD_NUM` be better? I can see it could at least avoid creating a separate tag for the generated image and likely save you from repeating the image name for gcr in some places.

.circleci/config.yml
```diff
@@ -33,26 +29,68 @@ jobs:
...
+          name: Update gcloud components
+          command: sudo /opt/google-cloud-sdk/bin/gcloud components update --quiet
+      - run:
+          name: boot cluster on gke
+          environment:
+            K8S_VERSION: 1.10.4-gke.2
+          command: gcloud container clusters create --cluster-version=$K8S_VERSION --disk-size=20 operator-test-$CIRCLE_BUILD_NUM
       - run:
           name: create image
           command: make TAG=testing image
```
krnowak: You could call `make IMAGE=${IMAGE_NAME} TAG=testing image` if you follow my suggestion for the `IMAGE_NAME` env var.

.circleci/config.yml
```diff
@@ -33,26 +29,68 @@ jobs:
...
+          name: Configure docker to use gcloud to authenticate requests to Container Registry
+          command: gcloud auth configure-docker
+      - run:
+          name: Tag the docker image
+          command: docker tag $IMAGE_NAME:testing eu.gcr.io/$GCLOUD_PROJECT_ID/$IMAGE_NAME-$CIRCLE_BUILD_NUM:testing
+      - run:
+          name: Push the docker image
+          environment:
+            GCLOUD_COMPUTE_ZONE: europe-west1-b
+          command: docker push eu.gcr.io/$GCLOUD_PROJECT_ID/$IMAGE_NAME-$CIRCLE_BUILD_NUM:testing
```
krnowak: You could use `${IMAGE_NAME}:testing` if you follow my suggestion about the `IMAGE_NAME` env var.

.circleci/config.yml
```diff
@@ -33,26 +29,68 @@ jobs:
...
+            K8S_VERSION: 1.10.4-gke.2
+          command: gcloud container clusters create --cluster-version=$K8S_VERSION --disk-size=20 operator-test-$CIRCLE_BUILD_NUM
       - run:
           name: create image
           command: make TAG=testing image
+      - run:
+          name: Configure docker to use gcloud to authenticate requests to Container Registry
+          command: gcloud auth configure-docker
+      - run:
+          name: Tag the docker image
```
krnowak: This step could be dropped if you follow my suggestion about the `IMAGE_NAME` env var.

.circleci/config.yml
```diff
@@ -33,26 +29,68 @@ jobs:
...
+            gcloud auth activate-service-account --key-file=${HOME}/gcloud-service-key.json
+            gcloud config set project $GCLOUD_PROJECT_ID
+            gcloud config set compute/zone $GCLOUD_COMPUTE_ZONE
+      - run:
+          name: Update gcloud components
+          command: sudo /opt/google-cloud-sdk/bin/gcloud components update --quiet
+      - run:
+          name: boot cluster on gke
+          environment:
+            K8S_VERSION: 1.10.4-gke.2
```
krnowak: Could you rename it to `GKE_K8S_VERSION`? I was a tiny bit confused if it doesn't clash with the `K8S_VERSION` env var from `setup` step.

.circleci/config.yml
```diff
@@ -33,26 +29,68 @@ jobs:
...
             until kubectl get nodes -o jsonpath="$JSONPATH" 2>&1 | grep -q "Ready=True"; do
               sleep 1
             done
+      - run:
+          name: Grant the user the ability to create authorization roles
+          command: kubectl create clusterrolebinding cluster-admin-binding --clusterrole cluster-admin --user $(gcloud config get-value account)
       - run:
           name: e2e tests
-          command: make TESTIMAGE=habitat/habitat-operator:testing e2e
+          command: make TESTIMAGE=eu.gcr.io/$GCLOUD_PROJECT_ID/$IMAGE_NAME-$CIRCLE_BUILD_NUM:testing e2e
```
krnowak: You could use `IMAGE_NAME` if you followed yadda yadda.

.circleci/config.yml
```diff
@@ -33,26 +29,68 @@ jobs:
...
           when: on_fail
-      # NOTE: this relies on journalctl, which is not present on the version of
-      # Ubuntu currently run by CircleCI
-      # - run:
-          # - name: print minikube logs
-          # - command: minikube logs
-          # - when: on_fail
+      - run:
+          name: Delete image from Container Registry
+          command: gcloud container images delete eu.gcr.io/$GCLOUD_PROJECT_ID/$IMAGE_NAME-$CIRCLE_BUILD_NUM:testing --force-delete-tags --quiet
```
krnowak: You could use `${IMAGE_NAME}:testing` if you followed my suggestion about the `IMAGE_NAME` env var.

.circleci/config.yml
kosyfrances: I did set it initially to `eu.gcr.io/$GCLOUD_PROJECT_ID/habitat/habitat-operator-$CIRCLE_BUILD_NUM` but the `$CIRCLE_BUILD_NUM` variable was not interpolated and I am not really sure why.

test/e2e/v1beta1/framework/helpers.go
```diff
@@ -263,3 +263,33 @@ func QueryService(url string) (string, error) {
...
+
+	if len(service.Status.LoadBalancer.Ingress) > 0 {
+		return service.Status.LoadBalancer.Ingress[0].IP, nil
+	}
+
+	return service.Spec.ClusterIP, nil
+}
+
+// WaitForLoadBalancerIP waits for Load Balancer IP to be available
+func (f *Framework) WaitForLoadBalancerIP(serviceName string) error {
```
krnowak: Why don't you just drop the `GetServiceIP` function and change the `WaitForLoadBalancerIP` function to return `(string, error)`? You could do it with something like:
```
address := ""
err := wait.Poll(…, func() (bool, error) {
    …
    if len(…) == 0 {
        return false, nil
    }

    address = service.Status.LoadBalancer.Ingress[0].IP
    return true, nil
}
return address, err
```

Maybe you will want to return `service.Spec.ClusterIP` if `wait.Poll` returns an error, but not sure how useful it is.

test/e2e/v1beta1/operator_test.go
```diff
@@ -100,10 +100,19 @@ func TestBind(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	// Wait until Load Balancer IP is ready
+	if err := framework.WaitForLoadBalancerIP(svc.ObjectMeta.Name); err != nil {
```
krnowak: You could move this call to happen after `time.Sleep(serviceStartupWaitTime)`, because right now we wait for the load balancer some time and then we unconditionally sleep for a fixed amount of time for the service to be ready. But the service is probably already ready by the time we get the load balancer IP. With the change, we can make the `TestBind` test a wee bit faster, because we will sleep for the service to be ready and maybe the load balancer will get the address at the same time and make the polling to last shorter.

.circleci/config.yml
```diff
@@ -33,26 +29,68 @@ jobs:
...
+      - run:
+          name: Configure docker to use gcloud to authenticate requests to Container Registry
+          command: gcloud auth configure-docker
+      - run:
+          name: Tag the docker image
+          command: docker tag $IMAGE_NAME:testing eu.gcr.io/$GCLOUD_PROJECT_ID/$IMAGE_NAME-$CIRCLE_BUILD_NUM:testing
+      - run:
+          name: Push the docker image
+          environment:
+            GCLOUD_COMPUTE_ZONE: europe-west1-b
```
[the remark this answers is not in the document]
kosyfrances: it has already been addressed ... 😕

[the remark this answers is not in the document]
krnowak: Yeah, in a separate place, where you dropped this env var in favor of just calling `gcloud config set compute/zone europe-west1-b` directly.

What I mean here is that this env var is not used in this step. The step is about pushing a docker image to some google docker registry and here the env var is not needed, right?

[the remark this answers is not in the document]
kosyfrances: the env var sets the default compute zone for the project. I just saw that I was setting it again where I was pushing the docker image so I will be taking that out instead and keeping `gcloud config set compute/zone europe-west1-b`

test/e2e/v1beta1/framework/helpers.go
kosyfrances: we can drop `service.Spec.ClusterIP` because the test will still fail if it does not have the `LoadBalancerIP` itself

kosyfrances:
@krnowak it is ready for another review :)

.circleci/config.yml
```diff
@@ -33,26 +33,61 @@ jobs:
...
+            echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" | sudo tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+            curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | sudo apt-key add -
+            sudo apt-get update && sudo apt-get install google-cloud-sdk
+
+            # the GCLOUD_SERVICE_KEY environment variable is set in the web UI
+            echo $GCLOUD_SERVICE_KEY > ${HOME}/gcloud-service-key.json
+
+            # setup SDK
+            gcloud auth activate-service-account --key-file=${HOME}/gcloud-service-key.json
+            gcloud config set project $GCLOUD_PROJECT_ID # GCLOUD_PROJECT_ID is set in Circle CI web UI
```
krnowak: Could you move this `# GCLOUD_PROJECT_ID is set …` comment to a separate line? Ideally above this one.

.circleci/config.yml
```diff
@@ -33,26 +33,61 @@ jobs:
...
+          command: sudo /opt/google-cloud-sdk/bin/gcloud components update --quiet
+      - run:
+          name: boot cluster on gke
+          environment:
+            GKE_CLUSTER_VERSION: 1.10.4-gke.2
+          command: gcloud container clusters create --cluster-version=$GKE_CLUSTER_VERSION --disk-size=20 operator-test-$CIRCLE_BUILD_NUM
       - run:
           name: create image
-          command: make TAG=testing image
+          command: make IMAGE=$IMAGE_NAME TAG=testing image
```
krnowak: `TAG=$IMAGE_TAG` please.

krnowak:
LFAD.

PR #303, opened 2018-06-29
Enable test for persistent storage

### LINKED ISSUE #292: Test PersistentVolumes
Blocked by #282.

kosyfrances:
We no longer run tests using minikube in a VM on the CI infrastructure
but with GCP

Fixes #292 

Depends on #302

### DIFF
--- a/test/e2e/v1beta1/operator_test.go
+++ b/test/e2e/v1beta1/operator_test.go
@@ -216,15 +216,12 @@ func TestHabitatDelete(t *testing.T) {
 }
 
 func TestPersistentStorage(t *testing.T) {
-	// We run minikube in a VM on the CI infrastructure. In that environment, we cannot create PersistentVolumes.
-	t.Skip("This test cannot be run successfully in our current testing setup")
-
 	ephemeral, err := utils.ConvertHabitat("resources/standalone/habitat.yml")
 	if err != nil {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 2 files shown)


krnowak:
Needs a rebase.

test/e2e/v1beta1/operator_test.go
```diff
@@ -255,6 +252,13 @@ func TestPersistentStorage(t *testing.T) {
 		}
 	})(persisted.Name)
 
+	// Delete the ephemeral resource created
```
krnowak: Should we also do the same for persistent one?

kosyfrances: Well, other tests do not use it so I skipped it instead. The whole cluster still gets deleted at the end of the test anyway.

kosyfrances:
```The test makes me wonder if we should validate that the pods are actually created. The creation of a pod can fail if the persistent volume claim can't be fulfilled. And I'm not sure if the tests actually tests it. What do you think?```

@krnowak I am not sure right now, let me think about it for a bit.

kosyfrances:
@krnowak The test actually waits for the persisted pod to be in a running state here https://github.com/habitat-sh/habitat-operator/pull/303/files#diff-2e3b6c005e3fc97b3b2a73f5cdeb0543R262

krnowak:
Oh, I overlooked that. In that case, LFAD.

Issue #238, opened 2018-04-11
Fix issues in scripts raised by shellcheck


krnowak:
Nothing serious, something to do if somebody suffers from too much free time.

```
[kv@localhost ([go-1.10, chef]) hack]$ shellcheck update-codegen.sh 

In update-codegen.sh line 29:
SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
                      ^-- SC2128: Expanding an array without an index only gives the first element.
                      ^-- SC2086: Double quote to prevent globbing and word splitting.


In update-codegen.sh line 30:
CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
                                ^-- SC2086: Double quote to prevent globbing and word splitting.


In update-codegen.sh line 36:
${CODEGEN_PKG}/generate-groups.sh deepcopy \
^-- SC2086: Double quote to prevent globbing and word splitting.


In update-codegen.sh line 39:
  --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.go.txt
                   ^-- SC2086: Double quote to prevent globbing and word splitting.

[kv@localhost ([go-1.10, chef]) hack]$ shellcheck verify-codegen.sh 

In verify-codegen.sh line 24:
SCRIPT_ROOT=$(dirname "${BASH_SOURCE}")/..
                       ^-- SC2128: Expanding an array without an index only gives the first element.

```

PR #306, opened 2018-07-05
Run e2e tests on multiple versions of kubernetes

### LINKED ISSUE #304: Run e2e tests in parallel with different k8s versions
Depends on #282 and #292

Add support for k8s 1.8 and 1.9 alongside 1.10.

kosyfrances:
Needs #303 to be merged first

Fixes #304 and #238 

This PR adds support for the tests to run successfully on k8s v1.8 and 1.9 alongside v1.10.

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -31,14 +31,14 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
-CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+CODEGEN_PKG=${CODEGEN_PKG:-$(cd "${SCRIPT_ROOT}"; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
 
 # generate the code with:
 # --output-base    because this script should also be able to run inside the vendor dir of
 #                  k8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir
 #                  instead of the $GOPATH directly. For normal projects this can be dropped.
-${CODEGEN_PKG}/generate-groups.sh all \
+"${CODEGEN_PKG}"/generate-groups.sh all \
   github.com/habitat-sh/habitat-operator/pkg/client github.com/habitat-sh/habitat-operator/pkg/apis \
   habitat:v1beta1 \
-  --go-header-file ${SCRIPT_ROOT}/hack/boilerplate.go.txt
+  --go-header-file "${SCRIPT_ROOT}"/hack/boilerplate.go.txt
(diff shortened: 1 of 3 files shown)


hack/update-codegen.sh
```diff
@@ -31,14 +31,20 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SCRIPT_ROOT=$(dirname ${BASH_SOURCE})/..
-CODEGEN_PKG=${CODEGEN_PKG:-$(cd ${SCRIPT_ROOT}; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[@]}")/..
```
krnowak: You should just pass the first item in `BASH_SOURCE` to `dirname`. So `${BASH_SOURCE[0]}`.

hack/update-codegen.sh
```diff
@@ -31,14 +31,20 @@ set -o errexit
...
+CODEGEN_PKG=${CODEGEN_PKG:-$(cd "${SCRIPT_ROOT}"; ls -d -1 ./vendor/k8s.io/code-generator 2>/dev/null || echo ../code-generator)}
 
 # generate the code with:
 # --output-base    because this script should also be able to run inside the vendor dir of
 #                  k8s.io/kubernetes. The output-base is needed for the generators to output into the vendor dir
 #                  instead of the $GOPATH directly. For normal projects this can be dropped.
-${CODEGEN_PKG}/generate-groups.sh all \
```
krnowak: I'm not sure this change is needed - we don't expect to run this script with some old kubernetes codegen version.

hack/verify-codegen.sh
```diff
@@ -23,7 +23,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-SCRIPT_ROOT=$(dirname "${BASH_SOURCE}")/..
+SCRIPT_ROOT=$(dirname "${BASH_SOURCE[@]}")/..
```
krnowak: Same as in the other script.

hack/verify-codegen.sh
```diff
@@ -32,10 +32,26 @@ ret=0
 
 git diff --quiet "*.deepcopy.go" || ret=$?
 
-if [[ $ret -eq 0 ]]
-then
-  echo "  Up to date."
+containsVersion () {
```
krnowak: I was thinking that we could skip the codegen checks in the circle ci config file instead of here. I mean, currently the codegen check step looks like this:
```
- run:
    name: code-gen script
    environment:
      …
    command: |
      mkdir -p …
      git clone … --branch $CODEGEN_VERSION …
      hack/verify-codegen.sh
```

What if we wrapped the command into `if [[ -n $CODEGEN_VERSION ]]; then mkdir …; git clone …; hack/verify-codegen.sh; fi`? And then drop the `CODEGEN_VERSION` variables from jobs for k8s 1.9 and 1.8?

That would basically give us just one place to edit (the circle ci config file) when we want to change the version of k8s the operator supports.

krnowak: You could probably add an `else` part to the `if` printing something about codegen checks being skipped or something.

krnowak:
Oh, and one more thing - you are fixing two issues in this PR - running tests for multiple k8s versions in parallel and addressing issues raised by shellcheck. Could you make two separate commits for each of those? They can still be in this PR.

kosyfrances:
@krnowak requested changes done!

krnowak:
LFAD.

Issue #345, opened 2018-08-22
Expose more StatefulSet+Pod+Container functionality through Habitat CRD


HT154:
The Habitat operator is great for running simple services, but for services with more advanced requirements, it can be rather limiting. Here are some features of Pods that the Habitat CRD should consider exposing:
* `affinity`
* `containers` - in addition to the primary operator-managed habitat-service container (e.g. running a Splunk forwarder)
* `dnsConfig`, `dnsPolicy`, `hostAliases`, `subdomain`
* `initContainers`
* `nodeSelector`, `tolerations`
* `securityContext` - particularly for setting `fsGroup: 42` (the hab user's gid) to fix persistent volume permissions issues
* `serviceAccountName` - potentially useful for running operators under the habitat operator
* `volumes` - in addition to the volumes configured by the operator (e.g. for sharing a logs directory with a Splunk forwarder sidecar)

Properties on the `habitat-service` Container that should be adjustable:
* `lifecycle`, `livenessProbe`, `readinessProbe`
* `resources` -  required for running in clusters that enforce pod resource limits/requests
* `securityContext`
* `volumeMounts` - in addition to operator-configured mounts, enables aforementioned Splunk use case

Properties on the StatefulSet that should be exposed:
* `podManagementPolicy`
* `serviceName`
* `volumeClaimTemplates` - in addition to operator-configured PVC templates

irvingpop:
Great list - the lack of liveness & readyness probes is a bummer.   I've also found myself wishing that the operator would expose topology type and leader election status.

HT154:
+1 On leader election status. It'd be great to set up a Service with a selector that selects for only the leader of a group. I was just wishing for this the other day for setting up redis replication.

HT154:
Updated original issue to remove StatefulSet `updateStrategy` from consideration. This is used internally by the operator to manage deleting pods and probably shouldn't be overridden.

clintoncwolfe:
This issue or pull request is being automatically closed because the repository is being archived as part of the [Repository Standardization Initiative](https://github.com/chef-boneyard/oss-repo-standardization-2025).

PR #351, opened 2018-09-05
Add ability to mount /files directory from a secret

jamesc:
This adds the ability to mount the `/files` directory of a habitat service via a secret much like is done for config.  What is different is that we add all the files that are part of the secret.

This is useful for services that e.g. load certificates via `hab file upload` right now but want to do it via a k8s primitive.

Signed-off-by: James Casey <james@chef.io>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ build:
 		github.com/$(GITHUB_ORG)/habitat-operator/cmd/habitat-operator
 
 .PHONY: linux
-linux:
+linux: build
 	# Compile statically linked binary for linux.
 	GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -ldflags="-s" \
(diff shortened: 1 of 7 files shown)


HT154:
I think this is better handled by implementing arbitrary volumes (+volumeMounts, see #345). If the operator supported configuring these in Habitat objects and piping them through to the underlying StatefulSet, it wouldn't be necessary for create bespoke implementations for every feature like this.

jamesc:
@HT154 I actually have a branch for the generic case too (https://github.com/jamesc/habitat-operator/tree/jc/generic-volume-mount) that I put together.  I went the more limited way as it fitted in with the current model better (similar to the configSecretName and ringSecretName parameters).

I agree there is a bunch of additional options that we need to have in the operator longer term (nodeSelector is the biggest problem for me right now). What I'm challenged by right now is a clean abstraction for adding them into the operator yaml without exposing the underlying primitives like Stateful Set/Pod/... to the end user like we do now.

